PR #32, opened 2023-03-05
`station metrics`. Closes #23

### LINKED ISSUE #23: `station metrics`

juliangruber:
Closes #23

### DIFF
--- a/bin/station.js
+++ b/bin/station.js
@@ -1,6 +1,12 @@
 #!/usr/bin/env node
 
 import * as commands from '../commands/index.js'
+import fs from 'node:fs/promises'
+import { join } from 'node:path'
+import { paths } from '../lib/paths.js'
+
+await fs.mkdir(join(paths.moduleStorage, 'saturn-L2-node'), { recursive: true })
+await fs.mkdir(paths.moduleLogs, { recursive: true })
 
 if (process.argv.includes('metrics')) {
   await commands.metrics()
--- a/bin/station.js
+++ b/bin/station.js
@@ -6,16 +6,12 @@ import fs from 'node:fs/promises'
 import * as saturnNode from '../lib/saturn-node.js'
 import { fileURLToPath } from 'node:url'
 import { createLogStream } from '../lib/log.js'
+import { Tail } from 'tail'
 
 const {
   FIL_WALLET_ADDRESS,
   XDG_STATE_HOME = join(homedir(), '.local', 'state')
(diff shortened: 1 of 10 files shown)


commands/metrics.js
```diff
@@ -0,0 +1,22 @@
+import fs from 'node:fs/promises'
+import { Tail } from 'tail'
+import { paths } from '../lib/paths.js'
+
+const metricsLogLineToJSON = (metrics) =>
+  JSON.stringify(
+    JSON.parse(metrics.split(/\[([^\]]+)\] /).pop()),
+    0,
+    2
+  )
```
juliangruber, line 10: I'm currently not sure about this part. We're storing metrics as log lines prefixed by `[$date] `, and now parsing them in order to output them as JSON. If we say that `$ station metrics` will be the primary way to look at metrics, then I'd propose to use NLDJSON, just as pino uses. While not as convenient when inspecting the files manually, it is easier to integrate and work with. Wdyt @bajtos?

bin/station.js
```diff
@@ -1,36 +1,15 @@
...
-  storagePath: join(paths.moduleStorage, 'saturn-L2-node'),
-  binariesPath: paths.moduleBinaries,
-  metricsStream: createLogStream(paths.metrics),
-  logStream: createLogStream(join(paths.moduleLogs, 'saturn-L2-node.log'))
-})
+if (process.argv.includes('metrics')) {
+  await commands.metrics()
+} else {
+  await commands.station()
+}
```
bajtos, line 15: Have you considered using a library to parse CLI args? Node.js is adding built-in support, but it's experimental, so we should probably use something else like minimise or yargs. 

https://oclif.io seems to be a more modern option.

bajtos (approved):
Nice! Kudos for adding automated tests 👏🏻

bin/station.js
juliangruber: I have thought about it, but for now that would be overkill imo. The simple logic here works and doesn't need (a tree of) dependencies. I'd suggest to add a lib like this (would be happy to try oclif, or the trusted minimist/yargs) once we see the current approach doesn't scale or causes bugs

commands/metrics.js
juliangruber: I agree 👍  I've opened a new issue here for final discussion: https://github.com/filecoin-station/core/issues/37

I believe we can ship this PR without it for now

PR #34, opened 2023-03-06
Sentry reporting. Closes #27

### LINKED ISSUE #27: Sentry reporting
- [ ] Add disclaimer to docs

juliangruber:
Closes #27.


bin/station.js
```diff
@@ -6,6 +6,17 @@ import fs from 'node:fs/promises'
...
+import * as Sentry from '@sentry/node'
+
+const repoRoot = join(dirname(fileURLToPath(import.meta.url)), '..')
+const pkg = JSON.parse(await fs.readFile(join(repoRoot, 'package.json')))
+
+Sentry.init({
+  dsn: 'https://6c96a5c2ffa5448d9ec8ddda90012bc9@o1408530.ingest.sentry.io/4504792315199488',
+  release: pkg.version,
+  environment: pkg.sentryEnvironment,
+  tracesSampleRate: 0.1
```
bajtos, line 18: Do we need to provide some sort of a tag to distinguish between data reported from the Desktop and the Core?

package.json
```diff
@@ -10,6 +10,8 @@
   },
   "scripts": {
     "test": "standard && node--test",
+    "preversion": "./scripts/pre-version.js",
+    "postpublish": "./scripts/post-publish.js",
```
bajtos, line 14: Hmm, this is a bit fragile. Do we have any documentation on the release process? IIUC, you are assuming we will always be using `npm version` followed by `npm publish`, right? 

It would be great to describe this aspect in the docs for the release process. If we ever decide to change it, e.g. automate it via GitHub Actions, then the docs should remind us about this aspect.

FWIW, pnpm has the concept of [`publishConfig`](https://pnpm.io/package_json#publishconfig). Unfortunately, it does not allow overwriting custom fields like `sentryEnvironment`.

bajtos (approved):

bin/station.js
juliangruber: It's a different dsn/project on Sentry. I think that's the easiest solution for now

package.json
juliangruber: We have documentation for it: https://github.com/filecoin-station/core#development If you `npm run release`, all of these steps are taken care of. But yes, this is assuming the release process runs `npm version` and then `npm publish`

PR #36, opened 2023-03-06
Docker. Closes #29

### LINKED ISSUE #29: Docker
- [x] Create docker container
- [x] Publish to docker registry
- [x] Document usage

```
docker run filecoin-station-core --name station --detach
```

- https://nodejs.org/en/docs/guides/nodejs-docker-webapp/
- https://www.docker.com/blog/how-to-use-the-alpine-docker-official-image/
- https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#non-root-user)

juliangruber:
Closes #29

https://github.com/filecoin-station/core/pkgs/container/core


juliangruber:
@bajtos the easy route is to publish to Docker once, a container that runs `npx ...`. The more laborious path is to publish to Docker in sync with NPM, where the docker image contains the real source. I assume this will also work better with watch tower. However, setting this up is more work. Can you please confirm the direction?

bajtos:
> @bajtos the easy route is to publish to Docker once, a container that runs `npx ...`.

Uh, oh. This IIUC means the Docker container will run a different version of Station Core code depending on when you start the container and the state of your local cache. When you restart the container, you may get a new version if Docker re-runs the layer with the `npx` command and downloads a fresh source code from `npm`, or run the cached version in other cases. This seems brittle and difficult to reason about to me.

As I understand Docker, images should be immutable. When I pull `postgres:14.5`, I always want to get PostgreSQL version 14.5 and not something newer.

For Station Core, I would like the Docker images to have the same versions as the `npm` package so there is a clear 1:1 mapping.

How much extra work is required? I assume we just need to add two or three docker commands to our release process (`build`, `tag`, `push`). It should be easy to write a script to automate this, shouldn't it?

bajtos (commented):
The changes LGTM to me. Could you please add the missing EOLs at EOFs?

Before we can land this, I think it's important to update the Developer documentation with instructions on how to publish new versions to Docker, possibly also how to join the `filecoin-station` team/org in Docker Hub.

bajtos:
GitHub offers documentation on how to setup GitHub Actions workflow to publish new versions to Docker, maybe we can use that? The idea is to create a workflow that runs on a new git tag and publishes to Docker.

https://docs.github.com/en/actions/publishing-packages/publishing-docker-images

bajtos:
This GitHub Actions looks easy to use:
https://github.com/marketplace/actions/build-and-push-docker-images

Here is an example showing how to configure multiple tags in addition to `latest`, e.g. `major` and `major.minor` and also how to publish Docker images from branches (pull requests) too.
https://docs.docker.com/build/ci/github-actions/manage-tags-labels/

Can we use that example configuration as the base and tweak it to our needs?

juliangruber:
@bajtos sgtm, thank you for requirement checking 🙏 

I'm going to add these commands and update docs. Thank you for sharing the resources, that's very helpful <3

juliangruber:
+1 to setting up publishing with GHA, so devs don't necessarily need to run docker and wait for things there

juliangruber:
https://github.com/filecoin-station/core/pkgs/container/core 🥳

juliangruber:
@bajtos I decided to go with just version tag and "latest" tag for now, let's revisit semver publishing if it comes up later?

.github/workflows/docker.yml
```diff
@@ -1,7 +1,7 @@
 name: Docker
 on:
   push:
-    tags: ["*"]
+    # tags: ["*"]
     paths:
       - ".github/workflows/docker.yml"
       - "Dockerfile"
```
bajtos: Ha, this is clever! I may want to use something similar to simplify debugging & troubleshooting the release workflow for Zinnia - https://github.com/filecoin-station/zinnia/pull/91

bajtos:
Can you please update installation instructions in README to show how to run Station Core via Docker?

juliangruber:
@bajtos this PR was already ready 👋

.github/workflows/docker.yml
juliangruber: I stole it from l1-node 👍

README.md
```diff
@@ -80,7 +80,18 @@ $ tail -f $XDG_STATE_HOME/filecoin-station/logs/modules/*.log
...
+
+Deploy Station with [Docker](https://www.docker.com/). Please replace
+`FIL_WALLET_ADDRESS`.
+
+```bash
+$ docker run \
+	-e FIL_WALLET_ADDRESS=f1abjxfbp274xpdqcpuaykwkfb43omjotacm2p3za \
+	-d ghcr.io/filecoin-station/core
+```
+
```
bajtos, line 93: Could you please add an example of obtaining metrics and list logs?

I think it would be something along the following lines:


```
docker exec ghcr.io/filecoin-station/core ./bin/station.js metrics
```

Docker docs: https://docs.docker.com/engine/reference/commandline/exec/

bajtos (approved):
👏🏻 LGTM with a few comments to consider. 

Feel free to address them in a follow-up pull request (or ignore them altogether).

.github/workflows/docker.yml
```diff
@@ -0,0 +1,47 @@
+name: Docker
```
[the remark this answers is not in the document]
juliangruber: Good idea!

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM node:18-alpine
+LABEL org.opencontainers.image.source https://github.com/filecoin-station/core
+USER node
+WORKDIR /usr/src/app
```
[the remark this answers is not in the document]
juliangruber: I believe it will be `~/.local/state/`, as that's the default `$XDG_STATE_HOME`. Do you think it should be somewhere else? Because if it should be somewhere else, we might consider also storing it somewhere else outside of Docker

README.md
juliangruber: great idea, done 🙏

Dockerfile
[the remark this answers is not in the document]
bajtos: > I believe it will be `~/.local/state/`, as that's the default `$XDG_STATE_HOME`.

Do we know the home dir (`~`) for our `node` user? 

Why I am asking: to inspect logs and other state files from outside, we need to know where to find those files inside the docker container.

Maybe we don't need that because users can run `station metrics` and `station logs` instead?

> Do you think it should be somewhere else? Because if it should be somewhere else, we might consider also storing it somewhere else outside of Docker

I was considering this too, but I don't know if storing the state outside of Docker is necessary.

I think the current setup is good enough and further improvements can wait until we get feedback from people running Station Core via Docker.

README.md
```diff
@@ -124,9 +124,11 @@ Deploy Station with [Docker](https://www.docker.com/). Please replace
 `FIL_WALLET_ADDRESS`.
 
 ```bash
-$ docker run \
-	-e FIL_WALLET_ADDRESS=f1abjxfbp274xpdqcpuaykwkfb43omjotacm2p3za \
-	-d ghcr.io/filecoin-station/core
+# Run container
+$ docker run --name station -d -e FIL_WALLET_ADDRESS=f1abjxfbp274xpdqcpuaykwkfb43omjotacm2p3za ghcr.io/filecoin-station/core
```
bajtos, line 128: Nitpick: I prefer the original version that splits the long line into multiple logical segments.

README.md
```diff
@@ -124,9 +124,11 @@ Deploy Station with [Docker](https://www.docker.com/). Please replace
...
+
+# Get logs
+$ docker exec station ./bin/station.js logs
```
bajtos, line 131: Let's not use shell comments and structure the text in Markdown instead.

GitHub does not seem to support a markdown snippet with embedded fenced code blocks; please replace two backticks with three backticks in the content below.


```md
Deploy Station with [Docker](https://www.docker.com/). Please replace 
`FIL_WALLET_ADDRESS`.

``bash
$ docker run (...)
``

Inspect logs:

``bash
$ docker exec station ./bin/station.js logs
``
```

bajtos (approved):
I have a couple more minor suggestions to consider.

Nothing major; the PR is good to be merged in the current form too.

Dockerfile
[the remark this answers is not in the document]
juliangruber: > Maybe we don't need that because users can run station metrics and station logs instead?

That's my thinking too, the user doesn't need to know

README.md
juliangruber: Good idea!

PR #39, opened 2023-03-08
fix stream errors not crashing the process

juliangruber:
Since this is a CLI, we want stream `"error"` events to crash the process, rather than just logging to stderr


bajtos (approved):
Fair enough 👍🏻

PR #40, opened 2023-03-09
fix flaky test

juliangruber:
Fixes https://github.com/filecoin-station/core/actions/runs/4373250537/jobs/7651184565

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -7,7 +7,6 @@ import { tmpdir } from 'node:os'
 import fs from 'node:fs/promises'
 import { randomUUID } from 'node:crypto'
 import { once } from 'node:events'
-import timers from 'node:timers/promises'
 import { getPaths } from '../lib/paths.js'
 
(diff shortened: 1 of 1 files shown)


test/test.js
```diff
@@ -50,17 +49,22 @@ test('Storage', async t => {
       XDG_STATE_HOME
     }
   })
-  await timers.setTimeout(1000)
+  while (true) {
```
bajtos, line 52: Have you seen this test fail? In other words, what happens if the file `saturn-L2-node.log` is never created? (You can try this failure mode by temporarily changing the filename in the test to an invalid name.)

Is my assumption correct that this never-ending loop will stop when the test runner aborts the test on a timeout?

juliangruber: Yes, it failed in the actions run linked above. The never ending loop will abort once the file exists

bajtos (approved):

PR #47, opened 2023-03-14
docs: update Getting started docs 

### LINKED ISSUE #25: Getting started documentation

bajtos:
- chore: setup Prettier formatting
- feat: specify supported Node.js versions
- docs: update Getting started docs

Resolve #25

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ $ FIL_WALLET_ADDRESS=f1... station
 ## Configuration
 
 - `$XDG_STATE_HOME` Station stores logs and module state in
-`$XDG_STATE_HOME/filecoin-station`. Defaults to `~/.local/state`.
+  `$XDG_STATE_HOME/filecoin-station`. Defaults to `~/.local/state`.
 
 ## Commands
 
@@ -38,10 +38,10 @@ Start Station.
 
 Configuration:
 
-- `$FIL_WALLET_ADDRESS` Filecoin wallet address (required). If you just want
-to give `core` a quick spin, you can use the address
-`f1abjxfbp274xpdqcpuaykwkfb43omjotacm2p3za`. Please note that any earnings
-sent there will be lost.
+- `$FIL_WALLET_ADDRESS` Filecoin wallet address (required). If you just want to
+  give `core` a quick spin, you can use the address
+  `f1abjxfbp274xpdqcpuaykwkfb43omjotacm2p3za`. Please note that any earnings
+  sent there will be lost.
 
 ### `$ station metrics`
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <h1 align="center">
 	<br>
-	 :artificial_satellite: 
+	 :artificial_satellite:
 	<br>
 	<br>
 	core
@@ -9,11 +9,22 @@
 	<br>
 </h1>
 
-> Repo for Filecoin Station Core
+> Station Core is a headless version of Filecoin Station suitable for running on
+> all kinds of servers.
 
 [![CI](https://github.com/filecoin-station/core/actions/workflows/ci.yml/badge.svg)](https://github.com/filecoin-station/core/actions/workflows/ci.yml)
 
+## Deployment
+
+Station Core supports different deployment options:
+
+- [Docker](#docker)
+- [Manual Deployment (Ubuntu)](#manual-deployment-ubuntu)
+
 ## Installation
+
+Station Core requires Node.js version 18.12 or newer. We recommend using the
+latest LTS version.
 
 ```bash
 $ npm install -g @filecoin-station/core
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -8,15 +8,19 @@
   "bin": {
     "station": "./bin/station.js"
   },
+  "engines": {
+    "node": ">=18.12"
```
bajtos, line 12: Hmm, I see that the CI is running on Node.js 16 too. Is there any particular reason why we should support that version line?

Either way, we should use the same minimal supported Node.js version in all three places - CI workflows, package.json, and documentation.

juliangruber: Node 16 has EOL 2023-09-11, so is still a relevant and supported version (https://github.com/nodejs/Release). Or would you like to choose different criteria?

juliangruber: Are you ok with supporting node 16? If yes, please adjust the other places to match

README.md
```diff
@@ -9,12 +9,23 @@
...
+## Deployment
+
+Station Core supports different deployment options:
+
+- [Docker](#docker)
+- [Manual Deployment (Ubuntu)](#manual-deployment-ubuntu)
+
 ## Installation
 
+Station Core requires Node.js version 18.12 or newer. We recommend using the
```
juliangruber, line 26: see Node.js version discussion

README.md
```diff
@@ -25,26 +36,40 @@ $ npm install -g @filecoin-station/core
...
-to give `core` a quick spin, you can use the address
-`f1abjxfbp274xpdqcpuaykwkfb43omjotacm2p3za`. Please note that any earnings
-sent there will be lost.
+  If you just want to give `core` a quick spin, you can use the address
+  `f1abjxfbp274xpdqcpuaykwkfb43omjotacm2p3za`. Please note that any earnings
+  sent there will be lost.
 
 ### `$ station metrics`
 
+_This command reads information from the state files created by the main Station
```
juliangruber, line 70: "main Station process" is a term not yet established. Also I don't think we need to talk about files here, does the user care where that data is coming from? To me this is too much an implementation detail. Therefore I wonder what you're trying to achieve by adding this text 🤔

README.md
```diff
@@ -64,17 +89,18 @@ $ station metrics --follow
 
 ### `$ station logs <module>`
 
-Get Station logs.
+_This command reads information from the state files created by the main Station
+process._
 
-Get all logs:
+Get a combined log from the Station Core and all Station Modules:
```
juliangruber, line 95:
```suggestion
Get combined logs from Station Core and all Station Modules:
```

juliangruber (changes requested):

package.json
bajtos: The fewer versions we support, the less maintenance work for us.

Considering that this is a CLI, not a library, I would assume people have more flexibility in choosing the Node.js version and thus be able to run on the latest LTS.

Anyhow, the goal of this PR is to add docs, so I'll update my changes to include support for Node.js 16.x.

bajtos: Resolved in https://github.com/filecoin-station/core/pull/47/commits/5cb8a745a1c020e995bfffc2b7c12acc118f806b

README.md
bajtos: Resolved in https://github.com/filecoin-station/core/pull/47/commits/5cb8a745a1c020e995bfffc2b7c12acc118f806b

README.md
bajtos: I am so glad you started this discussion, @juliangruber 🙏🏻

README.md
```diff
@@ -9,14 +9,25 @@
...
+- [Manual Deployment (Ubuntu)](#manual-deployment-ubuntu)
+
 ## Installation
 
+Station Core requires Node.js, it supports all maintained LTS major versions. We
+recommend using the latest LTS version.
+
 ```bash
-$ npm install -g @filecoin-station/core
+$ npm install -g --engine-strict @filecoin-station/core
```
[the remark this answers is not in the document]
juliangruber: Meh, that's ugly to me :| It should be the user's choice how they want to use NPM.

README.md
juliangruber: Hmm, if you compare this to `docker`, the `logs` command doesn't care if the container is running or not. I agree though that we should prevent concurrent `$ station` calls!

To me, annotating that logs are outdated doesn't make sense. The logs won't be outdated, they will be as uptodate as they can be. Only nothing is running that can produce new logs.

Can you maybe think of another CLI that works the way you propose?

juliangruber: > Maybe we can also improve all commands to print information about the location of state files at startup to make troubleshooting easier.

Are you sure this is a problem? Have you run into this?

juliangruber: > If you don't have the main Station Core process running, there will be either no metrics and no logs, or the data will be outdated.

To me it's obvious that there will be no new metrics or logs if the thing isn't running. But maybe I have a blind spot? It's the same with systemd: You can always get the logs. If you don't get new logs, you check if the service is running, and otherwise start it.

package.json
```diff
@@ -8,15 +8,19 @@
   "bin": {
     "station": "./bin/station.js"
   },
+  "engines": {
+    "node": "^16.13 || >=18.12"
```
juliangruber, line 12: why shouldn't someone on node 18.0 be able to run this? I don't see an advantage in forcing people to upgrade nodejs. Depending on the environment, they might not be able to (corporate policies etc), unnecessary limiting the possible install base of Station.
```suggestion
    "node": ">=16"
```

juliangruber (changes requested):

README.md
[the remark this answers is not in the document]
bajtos: 👍🏻

package.json
bajtos: The 16.x versions before 16.13 and 18.x versions before 18.12 were not LTS yet.

In the past, it was necessary to choose a non-zero server-minor version to get access to features added later in the lifecycle of the major version. E.g. top-level await was added in 14.8.

Anyhow, I don't have any particular argument or requirement to reject older 16.x and 18.x, so I am fine to drop the semver-minor part from the version spec.

However: your suggested change adds support for Node.js 17, which is no longer maintained. Are you fine with that? We don't run CI for that version.

bajtos: Also: do we want to automatically support all future Node.js versions by using `>=18`? Maybe the process of adding support for a new major version should be manual - add the new version to the CI matrix first, ensure all tests pass, only then update the version range.

README.md
```diff
@@ -9,12 +9,27 @@
...
+
+- [Docker](#docker)
+- [Manual Deployment (Ubuntu)](#manual-deployment-ubuntu)
+
 ## Installation
 
+> **Note**: Station Core requires Node.js, we recommend using the latest LTS
+> version. You can install Node.js using your favorite package manager or get
+> the official installer from
+> [Node.js downloads](https://nodejs.org/en/download/).
```
bajtos, line 29: Rendered version:

Screenshot 2023-03-15 at 14 36 13

See https://github.com/community/community/discussions/16925

bajtos:
Let's not bike shed this too much.

I think I addressed all your comments:
- Removed mentions of the main process.
- Removed specific Node.js versions from the README so that we don't have to remember to update README when we change which major versions are supported.
- Reworked engines spec to `>=16`.

Is there anything else needed before we can land this PR?

I would prefer to finish this ASAP, and then open follow-up pull requests if we want to make more changes.

README.md
juliangruber: This is looking awesome! 👏

PR #52, opened 2023-03-17
add lockfile

juliangruber:
This securely prevents multiple stations from running at the same time, as suggested by @bajtos 🙏 

Also as you suggested @bajtos, I refactored the `maybeCreateFile` function.


bajtos (approved):
👍🏻

PR #63, opened 2023-03-21
`GET /metrics`. Closes #55

### LINKED ISSUE #55: `GET /metrics?follow`

juliangruber:


README.md
```diff
@@ -142,7 +142,8 @@ Open HTTP API.
 
 Routes:
 
-- `/` 404 (Placeholder route)
+- `/` 404
```
bajtos, line 145: Should we remove this route? Should `/` return a simple HTML page listing the endpoints available?

lib/http.js
```diff
@@ -0,0 +1,45 @@
...
+      const controller = new AbortController()
+      const { signal } = controller
+      await Promise.all([
+        (async () => {
+          try {
+            await once(req, 'close')
+          } catch {
+          } finally {
+            controller.abort()
+          }
```
bajtos, line 20: This is a lot of code, mostly keywords and syntax tokens. Have you considered rewriting this using Promise API?

```js
once(req, 'close').then(
  _ok => controller.abort(),
  _err => controller.abort(),
);
````

File: lib/http.js [hunk 3]
```diff
@@ -0,0 +1,45 @@
...
+        })(),
+        (async () => {
+          try {
+            for await (const line of followMetrics({ signal })) {
+              res.write(`${line}\n`)
+            }
+          } catch {}
```
bajtos, line 27: I think you should end the response on error (e.g. when the `followMetrics` iterator throws). Otherwise, the response stream remains open forever.

lib/http.js
```diff
@@ -0,0 +1,45 @@
...
+        })()
+      ])
```
bajtos, line 29: I would find this code easier to understand if you extracted out the two inner functions.

```js
// here
      await Promise.all([
        abortOnRequestClosed(req, controller),
        pipeMetricsToResponse(res)
      ]);

// later - some of this is reusable!
        async function abortOnRequestClosed(req, controller) {
          try {
            await once(req, 'close')
          } catch {
          } finally {
            controller.abort()
          }
        }

        async function pipeMetricsToResponse(res, signal) {
          try {
            for await (const line of followMetrics({ signal })) {
              res.write(`${line}\n`)
            }
          } catch {
            // FIXME: close the response 
          }
        }
```

lib/http.js
```diff
@@ -0,0 +1,45 @@
+import { getLatestMetrics, followMetrics } from '../lib/metrics.js'
+import http from 'node:http'
+import { once } from 'node:events'
+
+const asyncHandler = async (req, res) => {
+  const url = new URL(req.url, 'http://localhost/')
+
+  if (url.pathname === '/metrics') {
```
bajtos, line 8: Since we want to add more endpoints soon, in order to avoid code churn, I am proposing to extract this handler into a standalone function, even if it feels like premature optimisation right now.

bajtos (commented):
Looks mostly good 👍🏻 

I have few comments to discuss though.

I'll take another & closer look tomorrow.

lib/http.js
juliangruber: I find this cleaner, as it's using the core language constructs `try` / `catch` and `finally` (proven from most languages), instead of the much debated Promise api. I understand it's more verbose

PR #65, opened 2023-03-22
ci: add macOS

juliangruber:


bajtos:
LGTM.

Considering the extra computing resources used, I wonder if we need to test all Node.js versions on non-Linux platforms. Maybe it's enough to test the latest Node.js LTS version on macOS and Windows? 

It makes the matrix definition slightly more complex, but not too much. I think something like this may work:

```yaml
    strategy:
       matrix:
         os: [ubuntu-latest]
         node: ['16', '18', '19']
         include:
           -  os: macos-latest
              node: 18
           -  os: windows-latest
              node: 18
 ```

juliangruber:
Some issues have only surfaced on certain OS combinations, like Windows 16 and mac 18. I'd prefer to keep it this way, unless we run into issues

PR #68, opened 2023-03-22
add passing tests for parallel commands

juliangruber:


test/test.js
```diff
@@ -136,17 +152,50 @@ test('Logs', async t => {
   })
 
   await t.test('Follow', async t => {
-    for (const flag of ['-f', '--follow']) {
+    await t.test('Read logs', async t => {
+      for (const flag of ['-f', '--follow']) {
```
bajtos, line 156: In my experience, it's better to move this for-loop outside of the test case, so that we have a new test for each variant we are checking. Such setup makes it easier to troubleshoot test failures, as they are more isolated.

```js
for (const flag of ['-f', '--follow']) {
  await t.test(`Read logs with "${flag}"`, async t => {
    // etc.
  })
}
```

bajtos (approved):
LGTM, thank you for adding these tests 👍🏻

test/test.js
juliangruber: This is a great suggestion! I knew something was off :D

PR #83, opened 2023-04-04
Move module bundling to `postinstall` script

juliangruber:
This way Station can bundle only the module binaries necessary for its operation.

Alternatives considered: Just like esbuild, have one package per os, all listed as optional dependencies:
- https://github.com/evanw/esbuild/blob/e9413cc4f7ab87263ea244a999c6fa1f1e34dc65/npm/esbuild/package.json
- https://github.com/evanw/esbuild/blob/e9413cc4f7ab87263ea244a999c6fa1f1e34dc65/npm/%40esbuild/android-arm/package.json

While this is safer (you use npm to download assets instead of a postinstall script), it's significantly more work. I propose to switch to that if the current approach becomes a problem.


bajtos (commented):
👍🏻

PR #92, opened 2023-04-12
feat: Zinnia Runtime and Peer Checker module

bajtos:
- feat: install Zinnia runtime

- feat: install mod-peer-checker

- feat: run peer checker module via Zinnia

- feat: switch base Docker image to full `node:18`

  >  We cannot use `node:18-alpine` because Zinnia does not support that yet.
  >
  >  I tried to use `node:18-slim`, but Saturn L2 was not able to start in that system.
  >  The Node.js project recommends to use `node:18`, so we should be fine. See https://github.com/nodejs/docker-node#image-variants

- test: move mocha config to config file and add `--exit`
 
- test: make `test/station.js` easier to troubleshoot
 
- test: fix expected Saturn messages

See https://github.com/filecoin-station/zinnia/issues/75

## TODO
- [x] Rebase on top of latest `main`
- [x] Unit & integration tests
- [x] Merge metrics from multiple modules
- [x] Docker
  - Add linux-arm64 builds to Zinnia
    https://github.com/filecoin-station/zinnia/pull/172
  - Rework Dockerfile to use Debian as the base image (Zinnia does not support Alpine yet)
  - Verify that Docker images work, especially on arm64
- [x] https://github.com/filecoin-station/zinnia/issues/148

## Out of Scope

- Access logs of individual Zinnia modules. It's tricky! I'll create a follow-up issue for that. 

- Documentation for Core developers
  - how to update Zinnia Runtime to a new version
  - how to update a Zinnia Module to a new version
  - how to add a new Zinnia Module to Station Core


lib/zinnia.js
```diff
@@ -0,0 +1,128 @@
...
+
+  const zinniadExe = join(moduleBinaries, `zinniad${platform() === 'win32' ? '.exe' : ''}`)
+  const modules = [
+    // all paths are relative to `moduleBinaries`
+    'peer-checker/peer-checker.js'
+  ]
+  const childProcess = execa(zinniadExe, modules, {
+    cwd: moduleBinaries,
+    env: {
+      WALLET_ADDRESS: FIL_WALLET_ADDRESS,
```
bajtos, line 34: TODO:
```suggestion
      FIL_WALLET_ADDRESS,
```

After we land & release this Zinnia PR:
- https://github.com/filecoin-station/zinnia/pull/179

juliangruber: Approved the PR 👍

bin/station.js
```diff
@@ -22,6 +22,8 @@ Sentry.init({
 
 await fs.mkdir(join(paths.moduleCache, 'saturn-L2-node'), { recursive: true })
 await fs.mkdir(join(paths.moduleState, 'saturn-L2-node'), { recursive: true })
+await fs.mkdir(join(paths.moduleCache, 'zinnia'), { recursive: true })
+await fs.mkdir(join(paths.moduleState, 'zinnia'), { recursive: true })
```
juliangruber, line 26:
```suggestion
for (const moduleName of ['saturn-L2-node', 'zinnia']) {
  await fs.mkdir(join(paths.moduleCache, moduleName), { recursive: true })
  await fs.mkdir(join(paths.moduleState, moduleName), { recursive: true })
}
```

I meant to refactor this once we add a second module

juliangruber: Could also pull this module list into a variable that can be reused in lines 46-51

scripts/post-install.js
```diff
@@ -92,3 +103,110 @@ if (target.archive === 'tar.gz') {
 }
 
 console.log(' ✓ %s', outFile)
+
+//= == Zinnia runtime ===//
```
juliangruber, line 107: Instead of adding structure comments, what do you think about moving the download logic into `lib/saturn.js` and `lib/zinnia.js`?

commands/station.js
```diff
@@ -27,13 +28,26 @@ export const station = async () => {
...
+      logStream: createLogStream(join(paths.moduleLogs, 'saturn-L2-node.log'))
+    }),
+
+    zinniaRuntime.start({
+      FIL_WALLET_ADDRESS,
+      STATE_ROOT: join(paths.moduleState, 'zinnia'),
+      CACHE_ROOT: join(paths.moduleCache, 'zinnia'),
+      moduleBinaries: paths.moduleBinaries,
+      // FIXME(bajtos) We need to merge metrics reported by different modules
+      metricsStream: createMetricsStream(paths.metrics.replace(/\.log$/, '-zinnia.log')),
```
juliangruber, line 48: This needs some refactoring from me: There should only be one metrics file, so the stats of all modules writing to it need to merged. Will work on it

juliangruber (changes requested):
over to me 😏

commands/station.js
juliangruber: Or rather: each module writes its own metrics file, and the `station metrics` command merges them

juliangruber: https://github.com/filecoin-station/core/pull/94

juliangruber: With that merged in, the current line can be changed to this:

```suggestion
      metricsStream: createMetricsStream('zinnia'),
```

juliangruber: Once `main` has been merged in, the snippet from the last comment should now work

scripts/post-install.js
bajtos: Yeah, we can definitely move this code. I wanted to avoid code churn, but I guess it doesn't matter that much.

Another thing to consider: if we move this code into new files, then we need to update our Dockerfile to copy those files over.

Since this is in `scripts/post-install.js`, I am thinking about introducing `scripts/lib/download-zinnia.js` and `scripts/lib/download-saturn.js`. WDYT?

commands/station.js
bajtos: Awesome, thank you!

scripts/post-install.js
juliangruber: Sounds all good to me! Docker should just copy over `script/*`

juliangruber: Can probably be refactored using https://github.com/filecoin-station/core/pull/104/files#diff-17a4b8942ea9f50b9ecadea71f1fc85944e98028b12e0220d6d45c38513329c3

juliangruber: If you think this lib will be useful, I can refactor it into a separate PR

bajtos:
It seems the Ubuntu Docker images are coming in minimal flavour by default, so hopefully there won't be too much bloat introduced by switching from Alpine to Debian 🤞🏻 

https://canonical.com/blog/minimal-ubuntu-released
> Minimal Ubuntu is a set of Ubuntu images designed for automated deployment at scale and made available across a range of cloud substrates. They use the optimised kernels and optimised boot process on their target compute substrate. These images have a greatly reduced default package set, without many convenience tools for interactive usage. They are much smaller, boot faster, and will require fewer security updates over time since they have fewer packages installed.

https://hub.docker.com/_/ubuntu
> Given that it is a minimal install of Ubuntu, this image only includes the C, C.UTF-8, and POSIX locales by default.

juliangruber:
> Any idea what may be wrong and/or how to troubleshoot?

I'll debug it on my Windows machine 👍  Basically for anything Windows specific you can defer that to me, as the team budget bought my Thinkpad.

juliangruber:
And if you need to sit at the machine, because it's more complicated etc, I'm happy to start the machine and give you remote access

juliangruber:
`.zip` extraction previously wasn't required on windows, so it didn't yet have logic for attaching the `.exe` extension. Should be fixed 👍

bin/station.js
```diff
@@ -21,6 +21,7 @@ Sentry.init({
 
 const core = new Core(getDefaultRootDirs())
 const modules = [
+  'runtime',
```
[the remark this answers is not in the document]
juliangruber: Same applies to the next file, can we call it `zinnia` instead of `runtime` everywhere?

lib/zinnia.js
```diff
@@ -0,0 +1,144 @@
...
+import { execa } from 'execa'
+import * as Sentry from '@sentry/node'
+import { installBinaryModule, downloadSourceFiles, getBinaryModuleExecutable } from './modules.js'
+import { moduleBinaries } from './paths.js'
+
+const ZINNIA_DIST_TAG = 'v0.7.0'
+const ZINNIA_MODULES = [
+  {
+    module: 'peer-checker',
+    repo: 'github.com/filecoin-station/mod-peer-checker',
```
juliangruber, line 11: The `downlaodSourceFiles` method depends on public GitHub APIs, so I think we can bake that assumption into the arguments by changing the repo string to just `filecoin-station/mod-peer-checker`

juliangruber (changes requested):

bajtos:
> `.zip` extraction previously wasn't required on windows, so it didn't yet have logic for attaching the `.exe` extension. Should be fixed 👍

Perfect, thank you for stepping in and fixing the issue!

Windows tests are passing on most versions, but there is a timeout on Node.js 16.

https://github.com/filecoin-station/core/actions/runs/4869576016/jobs/8684249782?pr=92#step:7:72
```
  1) Activity
       outputs activity:
     Error: Timeout of 15000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (D:\a\core\core\test\activity.js)
      at listOnTimeout (node:internal/timers:559:17)
      at processTimers (node:internal/timers:502:7)

  2) Activity
       can be read while station is running:
     Error: Timeout of 15000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (D:\a\core\core\test\activity.js)
      at listOnTimeout (node:internal/timers:559:17)
      at processTimers (node:internal/timers:502:7)
```

I'll re-run the test to see if this is intermittent. It would be nice to make our tests more robust, but I'd rather leave that out of this PR.

juliangruber:
Agreed, mostly I just re-run these but a better fix would be great

lib/zinnia.js
bajtos: Great idea. Addressed in f85c4a6

bin/station.js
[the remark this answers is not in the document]
bajtos: I initially started using `zinnia` everywhere but then decided to change the module name to `runtime`.

Here is my reasoning:

- In the longer term, we want all modules to be trusted and running via Zinnia. When that happens, we will have N+2 components in Station Core: Core itself, Module runtime (Zinnia) and N modules.
- As a Station user, I don't want to learn about Station internals and that we use Zinnia as the module runtime.
- As a Station user, when inspecting logs or activity, I can imagine what "runtime" might mean. I would have no idea what "zinnia" refers to.

> What do you think about calling the module `zinnia` instead? That's what I'd intuitively query

I think you have this intuition because you are a Station developer. Do you think Station users will find this intuitive too?

Anyhow, my opinion is not strongly held. If you think we should use `zinnia` even after considering my arguments, then let's do it.

From my perspective, it's important to build a consistent user experience. If we decide to use "Zinnia", we should use it in all user-facing places - the module name in the Activity log, the module name used for Core CLI commands like `core logs`, and so on.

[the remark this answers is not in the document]
juliangruber: Thank you for explaining your reasoning!

I think to me `runtime` is confusing since `bacalhau` also is a runtime, as `aqua` will be one. You're right that eventually we just want to have `core`, `runtime` and modules, but I see it taking a long time until the platform is both developed and rewarding enough (enough nodes) for zinnia to be the only option.

What do you think about calling it `zinnia` until it is the only runtime?

> From my perspective, it's important to build a consistent user experience. If we decide to use "Zinnia", we should use it in all user-facing places - the module name in the Activity log, the module name used for Core CLI commands like core logs, and so on.

+100, I'd call it `zinnia` _everywhere_, and nowhere `runtime`.

Wdyt?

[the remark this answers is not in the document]
bajtos: Fair enough, let's use `zinnia` everywhere 👌🏻

lib/zinnia.js
```diff
@@ -0,0 +1,144 @@
...
+              type: 'info',
+              message: event.message.replace(/Module Runtime/, 'Zinnia'),
+              source: event.module ?? 'Zinnia'
+            })
+            break
+
+          case 'activity:error':
+            activityStream.write({
+              type: 'error',
+              message: event.message.replace(/Module Runtime/, 'Zinnia'),
```
bajtos, line 126: Ditto.

lib/zinnia.js
```diff
@@ -0,0 +1,144 @@
...
+    .trimEnd()
+    .split(/\n/g)
+    .forEach(line => {
+      try {
+        const event = JSON.parse(line)
+        switch (event.type) {
+          case 'activity:info':
+            activityStream.write({
+              type: 'info',
+              message: event.message.replace(/Module Runtime/, 'Zinnia'),
```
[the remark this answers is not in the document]
juliangruber: WFM, since it's just these two calls here

juliangruber (approved):

bajtos:
@juliangruber I added three more commits to make the tests more robust, easier to troubleshoot and less prone to timing errors. PTAL at the changes and also let me know if I should extract those three commits into a standalone PR.

juliangruber (approved):

PR #100, opened 2023-04-20
Add API

juliangruber:
For https://github.com/filecoin-station/desktop/pull/621

- [x] Expose methods
- [x] Add typescript definition
- [x] Write docs
- [x] Add smoke tests
- [x] Remove types from git
- [x] Hide private methods


README.md
```diff
@@ -307,6 +307,57 @@ $ sudo systemctl start station
 $ sudo systemctl status station
 ```
 
+## API
+
+### `Core({ cacheRoot?: String, stateRoot?: String })`
+
+Returns `Core`.
```
bajtos, line 314: Let's make it more clear that `Core` is a class and the constructor must be called with the `new` keyword.

For example:

```suggestion
### `new Core({ cacheRoot?: String, stateRoot?: String })`

Creates a new instance of `Core`.
```

tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "include": ["index.js"],
+  "compilerOptions": {
+    "allowJs": true,
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    "outDir": "dist",
+    "declarationMap": true
+  }
+}
```
bajtos, line 10: Let's improve the compiler options a bit.

Here is what we have in Desktop:

```json
  "compilerOptions": {
    "noEmit": true,
    "strict": true,
    "skipLibCheck": true,
    "allowJs": true,
    "checkJs": true,
    "lib": ["es2021"],
    "target": "es2021",
    "module": "commonjs",
    "moduleResolution": "node",

    "resolveJsonModule": true,
    "allowSyntheticDefaultImports": true,
    "noFallthroughCasesInSwitch": true,
    "noImplicitReturns": true,
    "forceConsistentCasingInFileNames": true
  },
```

`skipLibCheck` is important to improve `tsc` speed.

`checkJs` is important to tell tsc to check our JS files and detect errors there too.

`lib` and `target` may be required to enable modern language features.

Personally, I tend to enable `strict`, because otherwise, it's easy to accidentally introduce `any` in the code, which then disables all type checks in the following code. The downside is that sometimes we need to describe types in more places, which is less ergonomic with JSDoc.

We need different `module` and `moduleResolution` values, since this is an ESM code base. I don't remember the correct values, though, TypeScript's support for ESM modules has been evolving quite a bit recently. See  https://www.typescriptlang.org/docs/handbook/esm-node.html

`forceConsistentCasingInFileNames` catches the problem when we use different letter casing in file names. macOS and Windows are not case-sensitive, but Linux is.

README.md
```diff
@@ -307,6 +307,57 @@ $ sudo systemctl start station
...
+
+Returns `Promise<String>`.
+
+### `Core#logs.follow(module?: String)`
+
+Returns `AsyncGenerator<String>`.
+
+### `Core#activity.get()`
+
+Returns `Promise<[ActivityEvent]>`.
```
bajtos, line 326: `[ActivityEvent]` means an array containing exactly on `ActivityEvent` item.

You should use `ActivityEvent[]` instead. (Alternatively, `Array<ActivityEvent>` should work too, but it's too verbose to my taste.)

bajtos (changes requested):
Great job!

The changes are heading in the right direction 👍🏻 

Let's improve the details.

tsconfig.json
bajtos: I opened #115 to address this thread.

bajtos:
I reviewed the `.d.ts` files produced by TypeScript. Many places contain `any`, presumably because the compiler did not understand what type is being used at runtime. I think enabling the strict mode would help us to find the causes.

Why is this a problem: when TypeScript encounters `any`, it stops doing any checks going further. With `any`, you can do whatever you want, and the compiler will never complain. You can pass a value of any type to a function argument accepting `any`. You can access arbitrarily-named properties and functions on the `any` value. Values returned by these properties and functions have `any` type too.

It would be great to fix these problems soon, but I guess they are not blocking this PR from being landed. However, replacing `any` with a more-specific type in our public API would be considered as a breaking change, so we may want to fix this before the next release.

Few examples of `any` in our public APIs:

`dist/lib/activity.d.ts`

```ts
export class Activity {
    constructor(core: any);
    core: any;
    // [snip]
    createWriteStream(source: any): AddMetaStream;
    // [snip]
}
```

`dist/lib/log.d.ts`

```ts
export function formatLog(text: any, date?: Date): string;
export function parseLog(line: any): {
    date: Date;
    text: any;
};
export class SerializeStream extends Transform {
    constructor();
    _transform(obj: any, _: any, callback: any): void;
}
export class Logs {
    constructor(core: any);
    core: any;
    // [snip]
    maybeCreateLogFile(module: any): Promise<void>;
    createWriteStream(path: any): PassThrough;
}
```

(and so on)

juliangruber:
> It would be great to fix these problems soon, but I guess they are not blocking this PR from being landed. However, replacing `any` with a more-specific type in our public API would be considered as a breaking change, so we may want to fix this before the next release.

There actually isn't any `any` in our public API, only in private APIs. The parts that you quoted aren't exposed directly, hence I think the risk is tolerable (for this stage of the project)

bajtos (approved):
Awesome! ✨

PR #104, opened 2023-04-24
Add `bacalhau`

juliangruber:


juliangruber:
Current state: Trying to figure out the metrics endpoint

Screenshot 2023-04-24 at 23 34 59

lib/bacalhau.js
```diff
@@ -0,0 +1,162 @@
...
+
+export async function install () {
+  await installBinaryModule({
+    module: 'bacalhau',
+    repo: 'bacalhau-project/bacalhau',
+    executable: 'bacalhau',
+    distTag: DIST_TAG,
+    targets: [
+      { platform: 'darwin', arch: 'x64', asset: `bacalhau_${DIST_TAG}_darwin_amd64.tar.gz` },
+      { platform: 'darwin', arch: 'arm64', asset: `bacalhau_${DIST_TAG}_darwin_arm64.tar.gz` },
```
bajtos, line 17: If they arm64 binary is not signed & notarised, then I won't be able to run it on my machine. We may need to use the same platform override as we do for Saturn L2.

I'll run this PR locally on my machine to see if it is a problem or not.

We can land this PR as-is; adding such an override can be easily done later.

lib/bacalhau.js
```diff
@@ -0,0 +1,162 @@
...
+      executable: 'bacalhau'
+    }),
+    [
+      'serve',
+      '--log-mode=station',
+      '--node-type=compute',
+      '--limit-total-cpu=1',
+      '--limit-total-gpu=0',
+      '--limit-total-memory=200Mb'
+      // '--disable-engine=docker' (not yet released)
```
bajtos, line 46: I think we should not land this PR until this new flag is released. WDYT?

bajtos (approved):
Very nice! 👏🏻

lib/bacalhau.js
juliangruber: It's still behind `--experimental`, I don't think this is a problem 🤔

lib/bacalhau.js
juliangruber: It did work for me though, and I'm on a M2

bajtos: This depends on your system's security settings too. 

- https://support.apple.com/en-us/HT202491
- https://gadgetstouse.com/blog/2021/02/05/run-unverified-unidentified-developer-apps-on-mac/#3_terminal_command

PR #107, opened 2023-04-26
ci: run tests for GB locale

bajtos:


commands/logs.js
```diff
@@ -1,11 +1,20 @@
...
+      const { text, date } = parseLog(line)
+      process.stdout.write(formatLog(text, { date, isHumanReadable: true }))
     }
   } else {
-    process.stdout.write(await getLatestLogs(module))
+    const lines = (await getLatestLogs(module))
+      .toString()
+      .trim()
+      .split('\n')
+      .filter(Boolean)
```
bajtos, line 14: Could you please remind me what this filter does? Are you filtering out empty lines?

lib/activity.js
```diff
@@ -30,7 +30,7 @@ export const createActivityStream = source => {
 const activityLogLineToObject = line => {
   const { date, text } = parseLog(line)
   const { type, message, source, id } = JSON.parse(text)
-  return { timestamp: date.toJSON(), type, source, message, id }
+  return { timestamp: date.toISOString(), type, source, message, id }
```
bajtos, line 33: I believe `Date.prototype.toJSON()` is using `toISOString()` under the hood, so this change is not necessary. 

Quoting from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toJSON:
> [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) instances refer to a specific point in time. toJSON() calls the object's [toISOString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) method, which returns a string representing the [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) object's value. This method is generally intended to, by default, usefully serialize [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) objects during [JSON](https://developer.mozilla.org/en-US/docs/Glossary/JSON) serialization, which can then be deserialized using the [Date() constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date) or [Date.parse()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse) as the reviver of [JSON.parse()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse).

In fact, why do we have to convert the date from `Date` to `String` here? I would expect the conversion to happen in the place that's converting the activity log line object to a string, as that's the place where we need to make the decision.

lib/activity.js
```diff
@@ -54,8 +54,8 @@ export const getActivity = async () => {
 export const maybeCreateActivityFile = () => maybeCreateFile(paths.activity)
 
 export const formatActivityObject = ({ type, message, timestamp }) => {
-  return formatLog(
-    `${type.toUpperCase().padEnd(5)} ${message}`,
-    new Date(timestamp)
-  )
+  return formatLog(`${type.toUpperCase().padEnd(5)} ${message}`, {
+    date: new Date(timestamp),
```
bajtos, line 58: Are we parsing `timestamp` from a `String` to a `Date` because we converted it from a `Date` to a `String` on line 33 above? Maybe we can avoid that roundtrip?

lib/log.js
```diff
@@ -8,15 +8,20 @@ import { Tail } from 'tail'
 import { platform } from 'node:os'
 import { on } from 'node:events'
 
-export const formatLog = (text, date = new Date()) =>
-  text
+export const formatLog = (
+  text,
+  { date = new Date(), isHumanReadable = false } = {}
```
bajtos, line 13: I find the name `isHumanReadable` a bit misleading since it does not tell us whether the value _is_ human readable now but whether we want to produce a human-readable representation in the output.

How about using just `humanReadable: true`?

Alternatively, what do you think about using `pretty: true` instead?

test/activity.js
```diff
@@ -28,14 +28,14 @@ describe('Activity', () => {
...
-      '[3/14/2023, 10:38:14 AM] {"source":"Saturn","type":"info","message":"beep boop"}\n'
+      '[2023-04-26T12:26:40.489Z] {"source":"Saturn","type":"info","message":"beep boop"}\n'
     )
     const { stdout } = await execa(
       station,
       ['activity'],
       { env: { CACHE_ROOT, STATE_ROOT } }
     )
-    assert.match(stdout, /3\/14\/2023/)
+    assert.match(stdout, /2023/)
```
bajtos, line 38: I guess this is good enough.

An alternative I used while trying to fix the failing tests - I am posting it just for consideration.

```suggestion
    assert.match(stdout, /(3\/14\/2023)|(14\/03\/2023)/)
```

bajtos (commented):
Great start! I have a few comments about the direction; let's discuss them.

commands/logs.js
bajtos: Cool! Would you mind adding a code comment to clarify this to readers unfamiliar with this trick?

test/activity.js
juliangruber: Then the next person changes their locale to German or something else and it will break again

lib/log.js
juliangruber: Went with `pretty`

lib/activity.js
juliangruber: I removed the conversion and changed it to be `timestamp` everywhere. I initially didn't do it because to me a property called `date` would be of type `Date`, any other type (like `String` or `Number` would be called something else. I realized now however that called a `Date` a `timestamp` makes sense. In my code before, a timestamp I think would always refer to a unix epoch.

lib/activity.js
juliangruber: fixed

commands/logs.js
juliangruber: Instead of using a trick, I would rather have something self explanatory. How would you write this part of the logic?

juliangruber:
@bajtos this is ready for another review 🙏

commands/logs.js
juliangruber: maybe like this?

```js
.filter(line => line !== '')
```

bajtos: I am fine with the trick if it's accompanied by a comment.

Your solution looks good to me too 👍🏻 

> ```js
> .filter(line => line !== '')
>```

test/activity.js
bajtos: I think there are only a handful of different ways to format a date, so I would not worry about scalability.

Anyhow, I am fine with checking just the year.

lib/activity.js
bajtos: Excellent! I find the new code much easier to comprehend 👍🏻

bajtos (commented):
@juliangruber LGTM 👏🏻 

I think you will need to approve this PR yourself, because it was opened by me 🙄

commands/logs.js
juliangruber: I don't want to use anything considered a "trick" because that makes the code harder to read, and if we also need to add a comment that makes it even more harder to read. The more explicit solution solves this elegantly.

juliangruber (approved):

PR #109, opened 2023-04-26
refactor `modules` library

juliangruber:
Pulled out of https://github.com/filecoin-station/core/pull/104/files and improved.


scripts/post-install.js
```diff
@@ -1,94 +1,3 @@
-#!/usr/bin/env node
```
bajtos, line 1: I think you need to preserve this line so that we can run `scripts/post-install.js` directly from the shell without prepending `node`.

lib/saturn-node.js
```diff
@@ -1,29 +1,42 @@
...
 import * as Sentry from '@sentry/node'
-
-const archOverwritten = platform() === 'darwin' ? 'x64' : arch()
+import { installBinaryModule, getBinaryModuleExecutable } from './modules.js'
+
+export async function install () {
+  await installBinaryModule({
+    module: 'saturn-l2-node',
+    repo: 'filecoin-saturn/L2-node',
+    distTag: 'v0.5.0',
```
bajtos, line 12: I slightly prefer to extract the dist tag into a named constant defined at the top of this source code file to make it easier to find.

bajtos (approved):
Great improvement!

The changes look mostly good, please consider my two comments below.

PR #91, opened 2023-04-12
Fix Docker build + setup a CI step

bajtos:
- build: change `npm ci` flags to `--omit=dev`
- build: copy post-install script to Docker
- ci: add a step to build the Docker image

Resolves #90

PR #110, opened 2023-04-26
ci: start Docker image and test output

bajtos:
- Fix Dockerfile to accommodate the latest changes and build a working Docker image again.
- Upgrade Saturn L2 Node to version [v0.5.1](https://github.com/filecoin-saturn/L2-node/releases/tag/v0.5.1)
- Add a new section to our CI workflow to build the Docker image, run it and check the activity log to verify that all modules have started.
- Partially revert 6bbd34c57fd7f07c5ce1324686ac9fd59f21c7a7 (#91). The "Docker" workflow is executed for release builds only; the pull requests are tested using the newly added section in the "CI" workflow.

More information on the users and permissions in the Node Docker images:
https://github.com/nodejs/docker-node/issues/740#issuecomment-458545074

Ideally, we would re-use the same workflow step to build the Docker image for release and for CI testing, just changing the output for the image(s). Unfortunately, that's not possible right now.
- `docker buildx` supports only single output, see https://github.com/moby/buildkit/issues/1555.
- The `docker` output type (used by `load: true` of the GitHub Action we are using) does not support multi-platform images, see https://docs.docker.com/engine/reference/commandline/buildx_build/#docker.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,13 +54,13 @@ jobs:
           cache-from: type=registry,ref=ghcr.io/filecoin-station/core
           cache-to: type=inline
 
-      - name: Build Docker image for other platforms
-        uses: docker/build-push-action@v4
-        with:
-          context: .
-          platforms: linux/arm64
-          cache-from: type=registry,ref=ghcr.io/filecoin-station/core
-          cache-to: type=inline
+      # - name: Build Docker image for other platforms
+      #   uses: docker/build-push-action@v4
+      #   with:
+      #     context: .
+      #     platforms: linux/arm64
+      #     cache-from: type=registry,ref=ghcr.io/filecoin-station/core
+      #     cache-to: type=inline
 
       - name: Start Station Core container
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -17,3 +17,65 @@ jobs:
...
+        env:
+          IMAGEID: ${{ steps.docker_build.outputs.imageid }}
+
+      - name: Print station activity
+        run: |
+          sleep 10 # Wait for Station modules to start
+          docker exec station bin/station.js activity
+
+      - name: Check | Saturn started
+        run: docker exec station bin/station.js activity | grep "Saturn module started"
```
bajtos, line 81: When we add a new module, we should add another step to check that the new module started too.

For example:

```
      - name: Check | Zinnia started
        run: docker exec station bin/station.js activity | grep "Module Runtime started"
```

@juliangruber FYI

juliangruber (approved):
Kickass!!!

PR #115, opened 2023-04-28
improve tsconfig + fix compiler errors

bajtos:
Let's improve our TypeScript setup.

I am not enabling `strict` mode yet, as that produces over 50 compiler errors that I don't have the bandwidth to investigate & fix right now.

I propose merging this PR into #100 first and looking into enabling the `strict` mode later.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,6 +2,17 @@
   "include": ["index.js"],
   "compilerOptions": {
     "allowJs": true,
+    "checkJs": true,
+    "skipLibCheck": true,
+    "lib": ["es2021"],
+    "target": "es2021",
+    "module": "Node16",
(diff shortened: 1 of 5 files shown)


tsconfig.json
```diff
@@ -2,6 +2,17 @@
   "include": ["index.js"],
   "compilerOptions": {
     "allowJs": true,
+    "checkJs": true,
+    "skipLibCheck": true,
+    "lib": ["es2021"],
+    "target": "es2021",
```
bajtos, line 8: I don't remember what ES version is supported by Node.js 16. We can downgrade these two lines to `es2020` if we run into runtime errors not caught by the compiler.

juliangruber (approved):
Thank you for stepping up!

PR #118, opened 2023-05-03
metrics: Fix optional module name api

juliangruber:


README.md
```diff
@@ -342,7 +342,7 @@ Returns `AsyncGenerator<ActivityEvent>`.
 
 Returns `Promise<MetricsEvent>`.
 
-### `Core#metrics.follow(module?: String, { signal?: AbortSignal })`
+### `Core#metrics.follow({ module?: String, signal?: AbortSignal })`
```
bajtos, line 345: Is it possible to omit both property args and call `follow` with no arguments at all?

```js
core.metrics.follow()
```

Maybe we can signal that option by adding `?` after the object.

```md
### `Core#metrics.follow({ module?: String, signal?: AbortSignal }?)`
```

Or perhaps:

```md
### `Core#metrics.follow({ module?: String, signal?: AbortSignal } = {})`
```

bajtos (approved):
👍🏻

README.md
juliangruber: Yes it's possible. I had the `= {}` before, but it got too verbose for my taste. Will try again

juliangruber:
https://github.com/filecoin-station/core/releases/tag/v8.0.0

PR #121, opened 2023-05-04
Add factory with setup

juliangruber:
Before this, only the CLI would set up the required files.

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,6 +15,7 @@ export class Core {
     'bacalhau'
   ]
 
+  /** @private */
   constructor ({ cacheRoot, stateRoot }) {
     this.paths = getPaths({ cacheRoot, stateRoot })
     this.logs = new Logs(this.paths.moduleLogs, this.paths.allLogs)


bajtos (commented):
Nice 👍🏻 

A potential problem I see is that users may forget to call `setup` before calling other methods. This can be solved by adding a guard to all methods to check that setup was already called. But that's cumbersome and requires us to remember to add this guard to every new method we add in the future.

A better alternative is to make the constructor private and provide a static factory function to create a new Core instance. This factory function can be async and create the necessary files.

To keep things simple, we can keep the constructor as a public undocumented function which can be used internally in this repo, and provide a public & documented static factory for consumers of this module.

```js
// before:
const core = new Core({/*...*/})
await core.setup()

// after
const score = await Core.create({/*...*/})
```

We can use a different name for `Core.create`, e.g. `Core.createClient`, `Core.setup`, `Core.initialize`, etc.

My preference: `Core.initialize`

index.js
```diff
@@ -22,4 +25,30 @@ export class Core {
       this.logs
     )
   }
+
+  async setup () {
```
bajtos, line 29: Should we make this method private?

```suggestion
  async #setup () {
```

bin/station.js
```diff
@@ -19,24 +19,7 @@ Sentry.init({
...
-await core.activity.maybeCreateActivityFile()
-await core.metrics.maybeCreateMetricsFile()
-await core.logs.maybeCreateLogFile()
-for (const module of modules) {
-  await fs.mkdir(join(core.paths.moduleCache, module), { recursive: true })
-  await fs.mkdir(join(core.paths.moduleState, module), { recursive: true })
-  await core.metrics.maybeCreateMetricsFile(module)
-  await core.logs.maybeCreateLogFile(module)
-}
+const core = await Core.create(getDefaultRootDirs())
```
bajtos, line 22: `Core.create` uses `getDefaultRootDirs` as the default argument value, so maybe we can simplify this line? Unless you want to be explicit about the directories.

```suggestion
const core = await Core.create()
```

index.js
```diff
@@ -2,17 +2,20 @@ import { Activity } from './lib/activity.js'
...
-   * @param {String} [options.stateRoot]
-   */
-  constructor ({ cacheRoot, stateRoot } = getDefaultRootDirs()) {
+  modules = [
+    'zinnia',
+    'saturn-L2-node',
+    'bacalhau'
+  ]
+
+  constructor ({ cacheRoot, stateRoot }) {
```
bajtos, line 18: Should we add a JSDoc comment to make the constructor private? We would need to inspect the `.d.ts` files produced by `tsc` to see if this has the desired effect of preventing people from calling the constructor directly.

```suggestion
  /** @private */
  constructor ({ cacheRoot, stateRoot }) {
```

test/api.js
```diff
@@ -3,34 +3,36 @@ import assert from 'node:assert'
...
+  describe('Core.create', () => {
+    it('can be constructed', async () => {
+      assert(await Core.create())
     })
-    it('can be constructed with custom paths', () => {
-      assert(new Core(getDefaultRootDirs()))
+    it('can be constructed with custom paths', async () => {
+      assert(await Core.create(getDefaultRootDirs()))
     })
   })
```
bajtos, line 18: Can you please add a test to verify that `Core.create` creates the necessary directories and files?

Ideally, such test should call Core API like "follow activity log" and describe a high-level scenario that we want to support, and it should fail when the setup is not called (i.e. without the changes you introduced here).

Considering our deadlines, we can land this PR with no new tests and add the tests in a follow-up PR, if you prefer.

bajtos (approved):

index.js
juliangruber: That works indeed! :)

test/api.js
juliangruber: I'm thinking the latter. Perfect!

PR #131, opened 2023-05-22
telemetry: add core id

juliangruber:
Add per process core random id. Without this, we can't tell how many core instances are running - only how many distinct wallets are configured.

A core process will get a random id when it starts. Since you can run multiple cores on one machine, a machine id isn't feasible.

### DIFF
--- a/lib/telemetry.js
+++ b/lib/telemetry.js
@@ -37,14 +37,14 @@ setInterval(() => {
 
 const startPingLoop = () => {
   assert(FIL_WALLET_ADDRESS)
-  const coreId = randomUUID()
+  const processUUID = randomUUID()
   return setInterval(() => {
     const point = new Point('ping')
     point.stringField(
       'wallet',
       createHash('sha256').update(FIL_WALLET_ADDRESS).digest('hex')
     )
-    point.stringField('core_id', coreId)
+    point.stringField('process_uuid', processUUID)
     point.stringField('version', pkg.version)
     point.tag('station', 'core')
     point.tag('platform', platform())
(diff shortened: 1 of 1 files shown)


lib/telemetry.js
```diff
@@ -37,12 +37,14 @@ setInterval(() => {
...
 const startPingLoop = () => {
   assert(FIL_WALLET_ADDRESS)
+  const coreId = randomUUID()
   return setInterval(() => {
     const point = new Point('ping')
     point.stringField(
       'wallet',
       createHash('sha256').update(FIL_WALLET_ADDRESS).digest('hex')
     )
+    point.stringField('core_id', coreId)
```
bajtos, line 47: I like the idea of adding a new field to determine how many unique Core instances are running at a given moment.

Using a unique id generated at process startup looks good to me too 👍🏻 

However, I have mixed feelings about the field name `core_id`. I would expect the id to be long-living and persistent across restarts, the same way how IDs are treated by e.g. databases or Docker daemon.

Can we find another name that will better convey the meaning?

I am thinking about the direction of "process id that's globally unique" or "unique instance id". Since this field is reported by Station Core, I feel we don't necessarily need to include `core_` prefix in the field name, as we already know the field is contributed by Station Core.

Few options to start our brainstorming:

- `process_uuid` 
- `instance_uuid`
- `running_instance_id`
- `instance_pid`

I feel `pid` (Process ID known from Unix) may be confusing because we will not report the actual PID reported by the operating system, but on the other hand, `instance_pid` is so succinct!

juliangruber: Great points!

I had also considered pids, but rejected them because they're not unique enough. And for that reason I also think we shouldn't use them in the field naming.

Out of your suggestions, I like `process_uuid` the most, since it conveys all the involved concepts

bajtos (approved):
:shipit:

juliangruber:
https://github.com/filecoin-station/core/releases/tag/v10.0.2

PR #134, opened 2023-05-24
remove storage (the great simplification)

juliangruber:


lib/activity.js
```diff
@@ -36,79 +18,18 @@ class ActivityEvent {
...
+const formatActivityObject = ({ type, message }) => {
+  return `${type.toUpperCase().padEnd(5)} ${message}`
+    .trimEnd()
+    .split(/\n/g)
+    .map(line => `[${(new Date()).toLocaleString()}] ${line}`)
+    .join('\n') + '\n'
 }
 
 module.exports = {
+  activities: new EventEmitter(),
```
juliangruber, line 32: @bajtos do you know what is best practice around typescript and event emitters? It would be cool to type this with `ActivityEvent`

lib/metrics.js
```diff
@@ -22,169 +14,56 @@ class MetricsEvent {
...
+        totalJobsCompleted: 0,
+        totalEarnings: 0
+      })
+    const isChanged = this.mergedMetrics === null ||
+      Object
+        .entries(this.mergedMetrics)
+        .some(([key, value]) => mergedMetrics[key] !== value)
+    if (isChanged) {
+      this.mergedMetrics = mergedMetrics
+      super.emit.call(this, 'update', mergedMetrics)
```
juliangruber, line 60: Question: This will emit metrics as soon as the first module emitted metrics. That means that while not all modules have emitted their first metrics event, we will show a partial job count. Should we delay emitting metric events until all modules have submitted? Risk: If one fails, we will never emit metrics. Also: I think this is something we can iterate on, and keep as is for now

lib/metrics.js
```diff
@@ -22,169 +14,56 @@ class MetricsEvent {
...
-    })
-    for await (const [line] of on(tail, 'line', { signal })) {
-      yield MetricsEvent.fromLogLine(line)
-    }
+class Metrics extends EventEmitter {
+  constructor () {
+    super()
+    this.mergedMetrics = null
+    this.moduleMetrics = new Map()
   }
```
juliangruber, line 24: I tried making `emit` uncallable:

```js
emit () {
  throw new Error('Use Metrics#submit')
}
```

Typescript complained that the method doesn't adhere to the required return type of `Boolean`. If I add a `return true` after the `throw`, the linter wants to remove it.

.github/workflows/ci.yml
```diff
@@ -89,10 +89,10 @@ jobs:
       - name: Print station activity
         run: |
           sleep 10 # Wait for Station modules to start
-          docker exec station bin/station.js activity
+          docker logs
```
bajtos, line 92: Should this perhaps be `docker logs station` for consistency with the other steps below?

README.md
```diff
@@ -273,12 +159,6 @@ $ docker run \
 	ghcr.io/filecoin-station/core
 ```
 
-Inspect logs:
-
-```bash
-$ docker exec station ./bin/station.js logs
-```
-
```
bajtos, line 281: How about we keep this section and change the command to `docker logs station`. WDYT?

README.md
```diff
@@ -316,57 +196,6 @@ $ sudo systemctl start station
 $ sudo systemctl status station
 ```
```
bajtos, line 198: Can we add a section describing how to inspect logs when running Station Core as a system service?

commands/station.js
```diff
@@ -3,96 +3,79 @@
...
+const { paths } = require('../lib/paths')
 
 const { FIL_WALLET_ADDRESS, MAX_DISK_SPACE } = process.env
 
-const station = async ({ core, json, experimental }) => {
+const moduleNames = [
+  'zinnia',
+  'saturn-L2-node',
+  'bacalhau'
+]
```
bajtos, line 19: I am surprised that you need to define a list of all valid modules. How comes we did not need this list before?

commands/station.js
```diff
@@ -3,96 +3,79 @@
...
-    console.error(`If you are sure this is not the case, please delete the lock file at "${core.paths.lockFile}" and try again.`)
-    process.exit(1)
+  startPingLoop().unref()
+  for (const moduleName of moduleNames) {
+    await fs.mkdir(join(paths.moduleCache, moduleName), { recursive: true })
+    await fs.mkdir(join(paths.moduleState, moduleName), { recursive: true })
   }
 
-  startPingLoop().unref()
+  activities.on('activity', activity => {
```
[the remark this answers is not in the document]
bajtos: 👌🏻

lib/activity.js
bajtos: In the past, I think I was using the solution described here: https://stackoverflow.com/a/39145058/69868

```ts
declare interface MyClass {
    on(event: 'hello', listener: (name: string) => void): this;
    on(event: string, listener: Function): this;
}

class MyClass extends events.EventEmitter {
    emitHello(name: string): void {
        this.emit('hello', name);
    }
}
```

Now we need to figure out how to do this with JSDoc comments. I'll give this a try and open a PR if I find a solution.

bajtos, line 32: I have mixed feelings about exposing a naked EventEmitter instance. It's super easy to implement, there is no unnecessary code. But: When reading the source code in this file, I have no idea what events will be emitted.

I prefer the approach in `lib/metrics.js`, where you export a class instance that provides methods for submitting new metrics data. That makes it much more clear how to work with the metrics module.

Would you mind using the same approach for activities too?

Instead of `submit`, maybe we can have `record` or `publish`? (Hmm, it's almost the same thing as `emit`, isn't it? 🤔)

lib/bacalhau.js
```diff
@@ -129,19 +119,18 @@ async function start ({
   } catch (err) {
     const errorMsg = err instanceof Error ? err.message : '' + err
     const message = `Cannot start Bacalhau: ${errorMsg}`
-    logStream.write(message)
-    activityStream.write({ type: 'error', message })
+    activities.emit('activity', { type: 'error', message })
```
bajtos, line 122: In the old version, activities were recorded in the logs in addition to being submitted to the activity stream. In the new version, they are reported in one place only. Could you please confirm that by emitting an activity this way, a log line will be printed to Core's stdout?

lib/metrics.js
bajtos: I think TypeScript can understand that your method never returns, therefore it infers the return type as `never` ([docs](https://www.typescriptlang.org/docs/handbook/basic-types.html#never)).

Would the following alternative work?

```js
/** @returns {Boolean} */
emit () {
  throw new Error('Use Metrics#submit')
}
```

bajtos: Alternatively, maybe we should not use inheritance. If we don't want Metrics to act as an event emitter, then we should not inherit from it.

```js
class Metrics {
  #emitter = new EventEmitter();
  // ...
  submit (moduleName, metrics) {
    // ...
    this.#emitter.emit(/*...*/);
  }

  onChange(fn) {
    this.#emitter.on('update', (metrics) => fn(metrics));
  }
}
```

This will also solve the problem of providing TypeScript typings for events. The event emitter is an internal implementation detail; thus it's kind of ok to have no types for it.

lib/metrics.js
```diff
@@ -22,169 +14,56 @@ class MetricsEvent {
...
-    assert(moduleName, 'moduleName required')
-    super({ objectMode: true })
-    this.lastMetrics = null
-    this.moduleName = moduleName
-  }
-
-  _transform (metrics, _, callback) {
-    if (typeof this.lastMetrics?.totalJobsCompleted === 'number') {
+  // - Merges metrics from all modules
+  submit (moduleName, metrics) {
```
bajtos, line 29: What is the type of `metrics`? Should it be `metricsEvent` with type `MetricsEvent`?

lib/metrics.js
bajtos: Let's iterate on this.

I suggest requiring all modules to emit the initial job count on startup. 
- We can handle this inside Core for modules exposing HTTP API, we just need to make the first `GET /stats` call immediately after the module starts accepting HTTP connections
- For modules running via Zinnia, we can easily improve `zinniad` to report the initial job count at startup - see https://github.com/filecoin-station/zinnia/blob/2b592f8ac636352ed3f1d6f43b771b5547496565/daemon/station_reporter.rs#L25-L40

lib/saturn-node.js
```diff
@@ -114,8 +101,8 @@ async function start ({
   childProcess.on('exit', (code, signal) => {
     const reason = signal ? `via signal ${signal}` : `with code: ${code}`
     const msg = `Saturn node exited ${reason}`
-    logStream.write(msg)
-    activityStream.write({ type: 'info', message: msg })
+    console.error(msg)
+    activities.emit('activity', { type: 'info', message: msg })
```
bajtos, line 105: I think it should be enough to emit an activity. We should not need to write the message to the console, since all activities are already logged. Thoughts?

lib/saturn-node.js
```diff
@@ -81,16 +69,15 @@ async function start ({
       if (apiMatch) {
         const apiUrl = apiMatch[1]
 
-        logStream.write('Saturn node is up and ready')
         childProcess.stdout.off('data', readyHandler)
-        activityStream.write({
+        activities.emit('activity', {
```
bajtos, line 73: ❗ 

I think you need to add a `source` field to tell which module is the source of this activity. 

**This comment applies to all other places in this file where you are calling `activities.emit`, and all other modules (Bacalhau, Zinnia) too.**

My suggestion: create a helper function replacing `activityStream.write` and setting the activity source.

```
// at the top start() function body
const logActivity = ({ type, message }) => {
  activities.emit('activity', { source: 'Saturn', type, message });
}

// here on line 73
logActivity({
  type: 'info',
  message: 'Saturn module started.'
})
```

lib/bacalhau.js
```diff
@@ -84,16 +75,15 @@ async function start ({
       if (apiMatch) {
         const apiUrl = apiMatch[1]
 
-        logStream.write('Bacalhau is up and ready')
         childProcess.stdout.off('data', readyHandler)
-        activityStream.write({
+        activities.emit('activity', {
           type: 'info',
           message: 'Bacalhau module started.'
         })
```
bajtos, line 82: Ditto - missing activity `source`

bajtos (changes requested):
I love how much simpler is the new code 😍 

The changes look mostly good 👍🏻 I have few minor comments that can be ignored or resolved after this PR is landed. 

However, I think there is one bug lurking. IIUC, you are not setting `source` field in the activities reported from modules. PTAL at the comment with a red exclamation mark.

lib/activity.js
bajtos: See https://github.com/filecoin-station/core/pull/134#discussion_r1203948272 - that may be the easiest solution.

bajtos: > I'll give this a try and open a PR if I find a solution.

I take that back. Let's look into this after we reach the final design and this PR is landed.

commands/station.js
juliangruber: we had it, only in `index.js`

lib/saturn-node.js
bajtos: This is what I see when I start Core with an invalid wallet address - notice the duplicated error line.

```
~/src/filstation/core on  the-great-simplification ❯ FIL_WALLET_ADDRESS=f1asd npm start

> @filecoin-station/core@10.0.2 start
> node ./bin/station.js

[24/05/2023, 14:20:59] INFO  Zinnia started.
Saturn node exited with code: 2
[24/05/2023, 14:20:59] INFO  Saturn node exited with code: 2
Saturn node closed all stdio with code 2
{
  "totalJobsCompleted": 0,
  "totalEarnings": 0
}
```

juliangruber: > I think it should be enough to emit an activity. We should not need to write the message to the console, since all activities are already logged. Thoughts?

Agreed, I just missed that one. Good catch!

lib/metrics.js
juliangruber: This helped fix a bug right away, done :)

lib/metrics.js
juliangruber: I don't think this is going to be good enough: If a module behaves badly, then it still won't report the metrics right away  (even if we pull, via the http interface).

README.md
bajtos: LOL, it reminds me of the old acronym [RTFM](https://en.wikipedia.org/wiki/RTFM). It wasn't a very friendly way how to treat users.

I agree documenting Docker CLI is out of scope of this README.

At the same time, I think a paragraph showing Station Core users how to access their logs has more benefits than costs.

Anyhow, I am fine either way. You decide.

lib/saturn-node.js
```diff
@@ -70,7 +70,8 @@ async function start ({
         const apiUrl = apiMatch[1]
 
         childProcess.stdout.off('data', readyHandler)
-        activities.emit('activity', {
+        activities.submit({
+          source: 'saturn-L2-node',
```
bajtos, line 74: ❗ 
I believe we are using human-friendly names, i.e.`Saturn`, `Zinnia`, `Bacalhau`. IIUC, they are directly rendered in Station Desktop UI. 

https://github.com/filecoin-station/core/blob/92858297ff092fd83d77f34c1e8371a3237a940b/commands/station.js#L37
https://github.com/filecoin-station/core/blob/92858297ff092fd83d77f34c1e8371a3237a940b/commands/station.js#L47
https://github.com/filecoin-station/core/blob/92858297ff092fd83d77f34c1e8371a3237a940b/commands/station.js#L59

Let's preserve this behaviour, please.

Nice to have: avoid repetition, extract the source string into a constant, or maybe introduce a per-module function for reporting activities that will add the right source value to all activities.

lib/metrics.js
```diff
@@ -22,169 +14,68 @@ class MetricsEvent {
...
-      'number',
-      'moduleMetrics.totalJobsCompleted required'
-    )
-    super({ objectMode: true })
-    if (!mergedMetrics) {
-      mergedMetrics = allMetrics
+    this.moduleMetrics.set(moduleName, metrics)
+    const mergedMetrics = Object
+      .values(this.moduleMetrics)
+      .reduce((acc, metrics) => {
```
bajtos, line 52: ❗ 
I think you have a bug in this code, which may also explain why I am not seeing any metric updates.

The first hint was offered by TypeScript, which was inferring both `acc` and `metrics` as `any`.

Apparently, you cannot use `Object.values()` on a map, you need to use `Map.prototype.values()` instead.

```
❯ node
Welcome to Node.js v18.14.2.
Type ".help" for more information.
> m = new Map();
Map(0) {}
> m.set("foo", "bar")
Map(1) { 'foo' => 'bar' }
> m
Map(1) { 'foo' => 'bar' }
> Object.values(m)
[]
> m.values()
[Map Iterator] { 'bar' }
```

Unfortunately, `[Map Iterator]` does not provide `map` method. An easy solution is to convert the iterator into a new array, paying the cost of extra allocation & copying:

```
> Array.from(m.values()).map(x => console.log(x))
bar
```

Personally, I would ditch `reduce` and replace it with a for-of loop:

```js
> for (x of m.values()) { console.log(x); }
bar
```

bajtos (changes requested):
Great improvements 👏🏻 Almost there!

lib/metrics.js
juliangruber: Great catch! Fixed using a for loop as suggested, it's also shorter this way

lib/saturn-node.js
juliangruber: Ok, I've refactored to the old behavior where `commands/station.js` owned the strings, now we're passing callbacks to the modules :)

commands/station.js
```diff
@@ -30,21 +30,19 @@ const station = async ({ json, experimental }) => {
...
 
-  activities.on('activity', activity => {
+  activities.onActivity(activity => {
     if (json) {
       console.log(JSON.stringify({
         type: `activity:${activity.type}`,
-        timestamp: activity.timestamp,
         module: activity.source,
-        message: activity.message,
-        id: activity.id
```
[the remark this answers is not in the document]
juliangruber: Since we don't persist to disk any more, the consumer (Station Desktop for example) can just attach `timestamp` and `id` while recording the events. Core doesn't need to care about these logistics I think

README.md
juliangruber: While I would never say RTFM, I also wouldn't say it's outdated - a lot of projects follow this practice I think. If you feel strongly about this, I'm fine to add the comment, but I personally value minimalism in code projects over friendliness - the friendly parts can make it harder to scan the documentation or code for what you're really looking for. A project's tone shouldn't be harsh and too dry either, but minimalistic neutral works very well for me.

lib/metrics.js
```diff
@@ -22,169 +14,65 @@ class MetricsEvent {
...
+    for (const [, metrics] of this.moduleMetrics) {
+      mergedMetrics.totalJobsCompleted += metrics.totalJobsCompleted
+    }
+    const isChanged = this.mergedMetrics === null ||
+      Object
+        .entries(this.mergedMetrics)
+        .some(([key, value]) => mergedMetrics[key] !== value)
+    if (isChanged) {
+      this.mergedMetrics = mergedMetrics
+      this.#events.emit('update', mergedMetrics)
```
bajtos, line 63: Can you please add a unit test or two for this code to make sure it really works as intended and does not break in the future?

commands/station.js
```diff
@@ -3,96 +3,89 @@
...
-      CACHE_ROOT: join(core.paths.moduleCache, 'zinnia'),
-      metricsStream: await core.metrics.createWriteStream('zinnia'),
-      activityStream: core.activity.createWriteStream('Zinnia'),
-      logStream: core.logs.createWriteStream(
-        join(core.paths.moduleLogs, 'zinnia.log')
-      )
+      STATE_ROOT: join(paths.moduleState, 'zinnia'),
+      CACHE_ROOT: join(paths.moduleCache, 'zinnia'),
+      onActivity: activity => {
+        activities.submit({ ...activity, source: activity.source || 'Zinnia' })
```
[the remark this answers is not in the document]
bajtos: Oh, `zinniad` is setting `source` to `null`, which does not play well with the spread operator. That explains why we need to use `source: activity.source || 'Zinnia'`.

Could you please add a code comment to remind future readers that we need to handle activities with `source: null`?

lib/metrics.js
bajtos: Sure. If we want to handle modules behaving badly, then we need a more robust solution.

I think my proposal is super easy to implement (and maintain) and will significantly improve the happy path. ATM, Zinnia does not update the job counter until the first job is finished, which takes around a second in my testing.

Anyhow, I think the current implementation is good enough for this pull request and we can iterate on improvements later.

bajtos (approved):
LGTM 👍🏻 

Please consider adding unit tests for the code merging metrics, per https://github.com/filecoin-station/core/pull/134#discussion_r1204598251

lib/metrics.js
bajtos: Opened a PR in Zinnia: https://github.com/filecoin-station/zinnia/pull/230

commands/station.js
```diff
@@ -3,96 +3,89 @@
...
-    await lockfile.lock(core.paths.lockFile)
-  } catch (err) {
-    console.error('Another Station is already running on this machine.')
-    console.error(`If you are sure this is not the case, please delete the lock file at "${core.paths.lockFile}" and try again.`)
-    process.exit(1)
+  startPingLoop().unref()
+  for (const moduleName of moduleNames) {
+    await fs.mkdir(join(paths.moduleCache, moduleName), { recursive: true })
+    await fs.mkdir(join(paths.moduleState, moduleName), { recursive: true })
   }
```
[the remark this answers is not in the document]
juliangruber: I thought about it, but I think this is an unnecessary optimization

PR #142, opened 2023-05-31
remove `peer-checker`, add `spark`

juliangruber:
Thank you `peer-checker`, you have served well 👋

### DIFF
--- a/lib/zinnia.js
+++ b/lib/zinnia.js
@@ -7,9 +7,9 @@ import { moduleBinaries } from './paths.js'
 const ZINNIA_DIST_TAG = 'v0.10.1'
 const ZINNIA_MODULES = [
   {
-    module: 'peer-checker',
-    repo: 'filecoin-station/mod-peer-checker',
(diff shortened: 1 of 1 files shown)


bajtos (approved):
I am excited to see the first real Zinnia module coming to Station! 🕺🏻 🎉 

The changes LGTM :shipit: 

Two more thoughts:
- After we land this PR, let's archive the GitHub repo for peer-checker, to make it clear that we are not going to maintain it going forward.
- I am wondering how is `zinniad` going to handle the removal of a module. Will it preserve job count? Will there be any directories left behind for the old module? I think all should be good, but it's worth taking a look.

juliangruber:
> I am wondering how is zinniad going to handle the removal of a module. Will it preserve job count? Will there be any directories left behind for the old module? I think all should be good, but it's worth taking a look.

I haven't noticed anything out of the ordinary 👍

juliangruber:
https://github.com/filecoin-station/core/releases/tag/v12.0.0

PR #150, opened 2023-06-06
telemetry: add `ARCH`

juliangruber:


bajtos:
Let's try the solution based on spawning `sysctl sysctl.proc_translated` first so that we avoid this new env var.

juliangruber:
https://github.com/filecoin-station/desktop/pull/767

PR #154, opened 2023-06-14
add `$ARCH` to install script

juliangruber:
Required for https://github.com/filecoin-station/desktop/pull/767

### DIFF
--- a/lib/modules.js
+++ b/lib/modules.js
@@ -13,7 +13,7 @@ import { moduleBinaries } from './paths.js'
 
 /** @typedef {import('unzip-stream').UnzipStreamEntry} UnzipStreamEntry */
 
-const { GITHUB_TOKEN, ARCH = os.arch() } = process.env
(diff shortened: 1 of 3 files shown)


lib/modules.js
```diff
@@ -36,21 +36,20 @@ export const installBinaryModule = async ({
   repo,
   distTag,
   executable,
-  arch = os.arch(),
```
juliangruber, line 39: this argument wasn't being used any more

bajtos: I am concerned about setting the arch deep inside this file. When reading the top-level post-install script, it's not clear that we may be downloading different files based on some environment variables.

Would you mind keeping `arch` here, maybe as a required argument, and setting it from env vars in the file where we are calling `getBinaryModuleExecutable`?

juliangruber: good idea, like this?

PR #161, opened 2023-06-20
Fix unzip on Windows

juliangruber:
I don't know why, but the old flow made node crash with code 13, no error message

### DIFF
--- a/lib/modules.js
+++ b/lib/modules.js
@@ -7,7 +7,6 @@ import { pipeline } from 'node:stream/promises'
 import gunzip from 'gunzip-maybe'
 import tar from 'tar-fs'
 import unzip from 'unzip-stream'
-import { once } from 'node:events'
 import { createWriteStream } from 'node:fs'
 import { moduleBinaries } from './paths.js'
 
--- a/lib/modules.js
+++ b/lib/modules.js
@@ -79,23 +79,15 @@ export const installBinaryModule = async ({
   } else {
     await mkdir(join(moduleBinaries, module), { recursive: true })
     const parser = unzip.Parse()
-    await Promise.all([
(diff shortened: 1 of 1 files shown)


bajtos (commented):
I have mixed feelings about mixing callback-style events and async functions. IIRC, when the async event listener function fails, there is nobody to handle the rejected promise. Having said that, as long as your solution works, I am fine with it.

juliangruber:
I don't like it either. Can you suggest a way to improve this?

To me the async function in the event emitter listener is ok, because if it fails it will still make the process crash

bajtos:
> To me the async function in the event emitter listener is ok, because if it fails it will still make the process crash

Yeah, makes sense; I think this is actually the simplest option. Maybe we can add a code comment to explicitly explain that we rely on that behaviour for error handling. So that people reading the code know that we are intentionally using this solution for error handling instead of accidentally not considering error handling at all.

A more explicit solution is to wrap everything in a promise, but it's complex and ugly, and not that much better in this particular case where we want to crash the process anyways.

```js
await Promise.all([
  () => new Promise((resolve, reject) => { 
    parser.on('entry', async entry => {
      try {
        const executableFileName = getExecutableFileName(executable)
        const outPath = join(moduleBinaries, module, entry.path)
        await pipeline(entry, createWriteStream(outPath))
        if (entry.path === executableFileName) {
          await chmod(outPath, 0o755)
        }
      } catch (err) {
        reject(err)
      }
    })
    // When to resolve the promise? Does the parser tell us when it's done?
    parser.on('end', () => resolve());
  }),
  pipeline(res.body, parser)
])
```

Let's keep your current fix 👍🏻

PR #169, opened 2023-06-29
fix: restart Zinnia after it exits

bajtos:
WIP

Related: https://github.com/filecoin-station/zinnia/issues/267

PR #170, opened 2023-06-29
add restart modules

juliangruber:
Alternative implementation to https://github.com/filecoin-station/core/pull/169.

Notable differences:

Instead of the zinnia lib restarting itself internally, we make `zinnia.start()` throw an error when it exists. Then, we use `p-retry()` to take care of restarting it

### DIFF
--- a/commands/station.js
+++ b/commands/station.js
@@ -51,7 +51,7 @@ export const station = async ({ json, experimental }) => {
   })
 
   const modules = [
-    pRetry(() => zinniaRuntime.start({
+    pRetry(() => zinniaRuntime.run({
       FIL_WALLET_ADDRESS,
       STATE_ROOT: join(paths.moduleState, 'zinnia'),
       CACHE_ROOT: join(paths.moduleCache, 'zinnia'),
--- a/commands/station.js
+++ b/commands/station.js
@@ -6,6 +6,7 @@ import * as bacalhau from '../lib/bacalhau.js'
 import fs from 'node:fs/promises'
 import { metrics } from '../lib/metrics.js'
 import { paths } from '../lib/paths.js'
+import pRetry from 'p-retry'
 
 const { FIL_WALLET_ADDRESS } = process.env
 
@@ -50,7 +51,7 @@ export const station = async ({ json, experimental }) => {
   })
 
   const modules = [
-    zinniaRuntime.start({
+    pRetry(() => zinniaRuntime.start({
       FIL_WALLET_ADDRESS,
       STATE_ROOT: join(paths.moduleState, 'zinnia'),
       CACHE_ROOT: join(paths.moduleCache, 'zinnia'),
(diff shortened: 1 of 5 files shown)


commands/station.js
```diff
@@ -64,7 +65,7 @@ export const station = async ({ json, experimental }) => {
         })
       },
       onMetrics: m => metrics.submit('zinnia', m)
-    })
+    }), { retries: 1000 })
```
juliangruber, line 68: I tried `Infinity`, unfortunately that doesn't work. See https://github.com/sindresorhus/p-retry/issues/58

bajtos: Can we use something like a max-safe integer?

commands/station.js
```diff
@@ -50,7 +51,7 @@ export const station = async ({ json, experimental }) => {
   })
 
   const modules = [
-    zinniaRuntime.start({
+    pRetry(() => zinniaRuntime.start({
```
bajtos, line 54: With this new implementation, I think the name `start` is no longer accurate. What do you think about renaming `start` to `run`?

```suggestion
    pRetry(() => zinniaRuntime.run({
```

juliangruber: Good call, yes!

lib/zinnia.js
```diff
@@ -72,29 +79,30 @@ export async function start ({
...
-  } catch (err) {
-    const errorMsg = err instanceof Error ? err.message : '' + err
-    const message = `Cannot start Zinnia: ${errorMsg}`
-    onActivity({ type: 'error', message })
-  }
+  await Promise.all([
+    (async () => {
+      try {
+        await Promise.race([
+          readyPromise,
```
juliangruber, line 86: remove `readyPromise`

commands/station.js
juliangruber, line 1: do the same thing for `bacalhau`

lib/zinnia.js
```diff
@@ -57,6 +58,12 @@ export async function start ({
     }
   })
 
+  childProcess.on('exit', (code, signal) => {
```
juliangruber, line 61: move this back below

lib/zinnia.js
juliangruber:
```ja
await childProcess
throw new Error()
```

commands/station.js
juliangruber: it also fails with that unfortunately

bajtos (approved):

lib/zinnia.js
```diff
@@ -57,44 +57,38 @@ export async function start ({
...
-    childProcess.stdout.on('data', data => {
-      handleEvents({ onActivity, onMetrics, text: data })
-    })
-    childProcess.stderr.pipe(process.stderr, { end: false })
-
-    childProcess.stdout.once('data', _data => {
-      // This is based on an implicit assumption that zinniad reports an info activity
-      // after it starts
-      resolve()
-    })
```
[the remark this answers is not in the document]
juliangruber: That is correct. It was the outcome of a meeting by the two of us that this logic can be dropped. I agree, let's see how this goes

PR #223, opened 2023-10-02
remove `undici`

juliangruber:
This is a semver major change, because it drops Node 16 support

### DIFF
--- a/lib/modules.js
+++ b/lib/modules.js
@@ -2,13 +2,13 @@ import os from 'node:os'
 import assert from 'node:assert'
 import { join } from 'node:path'
 import { mkdir, chmod } from 'node:fs/promises'
-import { fetch } from 'undici'
 import { pipeline } from 'node:stream/promises'
 import gunzip from 'gunzip-maybe'
 import tar from 'tar-fs'
 import unzip from 'unzip-stream'
 import { createWriteStream } from 'node:fs'
 import { moduleBinaries } from './paths.js'
+import { streamAsyncIterable } from './util.js'
 
 /** @typedef {import('unzip-stream').UnzipStreamEntry} UnzipStreamEntry */
 
@@ -74,7 +74,7 @@ export const installBinaryModule = async ({
   }
 
   if (target.asset.endsWith('tar.gz')) {
-    await pipeline(res.body, gunzip(), tar.extract(outFile))
+    await pipeline(streamAsyncIterable(res.body), gunzip(), tar.extract(outFile))
   } else {
     await mkdir(join(moduleBinaries, module), { recursive: true })
     const parser = unzip.Parse()
@@ -86,7 +86,7 @@ export const installBinaryModule = async ({
         await chmod(outPath, 0o755)
       }
     })
-    await pipeline(res.body, parser)
+    await pipeline(streamAsyncIterable(res.body), parser)
   }
   console.log(`[${module}] ✓ ${outFile}`)
 }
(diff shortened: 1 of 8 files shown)


lib/modules.js
```diff
@@ -74,7 +74,7 @@ export const installBinaryModule = async ({
   }
 
   if (target.asset.endsWith('tar.gz')) {
-    await pipeline(res.body, gunzip(), tar.extract(outFile))
+    await pipeline(streamAsyncIterable(res.body), gunzip(), tar.extract(outFile))
```
juliangruber, line 77: I feel like I'm doing something wrong here. However, MDN suggests this approach. I didn't find anything in node's stdlib for this 🤔

tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "allowJs": true,
     "checkJs": true,
     "skipLibCheck": true,
-    "lib": ["es2021"],
+    "lib": ["es2021", "DOM"],
```
juliangruber, line 6: is there a better lib target to add? See https://www.typescriptlang.org/tsconfig#lib. `DOM` was suggested in stack overflow posts

bajtos: Oh man, this is a rabbit hole. See the discussion in https://github.com/DefinitelyTyped/DefinitelyTyped/issues/60924.

First of all, Station Core is currently supporting Node.js `16.x` which does not provide a built-in Fetch API. So the first step is to upgrade our minimal supported version to 18.x - I think this is a breaking change.

Then, ideally, all we need to do is add `@types/node@18` to our dev-dependencies.

However, that does not work because of the reasons described in https://github.com/DefinitelyTyped/DefinitelyTyped/issues/60924.

Also, we should not be using DOM types, because Fetch API on server has slightly different API than in the browser. (This was mentioned in that DefinitelyTyped discussion too.)

I quite like the workaround described https://github.com/DefinitelyTyped/DefinitelyTyped/issues/60924#issuecomment-1563615437. The downside is that we still need to have undici in our dependencies (and probably keep it up to date via Dependabot?), but at least it becomes a dev-time only dependency. 🤷🏻‍♂️ 

```
// fetch.d.ts
import type * as undici from "undici";

declare global {
  // Re-export undici fetch function and various classes to global scope.
  // These are classes and functions expected to be at global scope according to
  // Node.js v18 API documentation.
  // See: https://nodejs.org/dist/latest-v18.x/docs/api/globals.html
  export const { 
    FormData,
    Headers,
    Request,
    Response,
    fetch,
  }: typeof undici;
}
```

bajtos: Here is a diff of the changes I think we should make on our way to fix this properly. Unfortunately, it's not enough. 

```diff
diff --git a/package.json b/package.json
index edc5522..abe92f3 100644
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "station": "./bin/station.js"
   },
   "engines": {
-    "node": ">=16"
+    "node": ">=18"
   },
   "scripts": {
     "build": "tsc -p .",
@@ -32,6 +32,7 @@
     ]
   },
   "devDependencies": {
+    "@types/node": "^18.18.1",
     "@types/yargs": "^17.0.24",
     "get-stream": "^8.0.1",
     "mocha": "^10.2.0",
diff --git a/tsconfig.json b/tsconfig.json
index 5d96c50..83db0fb 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,7 +3,7 @@
     "allowJs": true,
     "checkJs": true,
     "skipLibCheck": true,
-    "lib": ["es2021", "DOM"],
+    "lib": ["es2021"],
     "target": "es2021",
     "module": "Node16",
     "moduleResolution": "node16",
```

juliangruber: Thank you for doing the research! I've applied the fixes you suggested. I don't think overall this improves our situation unfortunately, and I'd rather propose to continue using undici for now

bajtos: > I don't think overall this improves our situation unfortunately, and I'd rather propose to continue using undici for now

💯 agree

package.json
```diff
@@ -32,14 +32,16 @@
     ]
   },
   "devDependencies": {
+    "@types/node": "^20.8.0",
```
bajtos, line 35: FWIW, this should be `^18.x.y` (for the latest x and y). Otherwise, the TypeScript compiler will allow us to use APIs introduced in Node.js 20 that are not available in Node.js 18 yet.

PR #259, opened 2023-11-10
Print `scheduledRewards` instead of `totalEarnings`. Closes #255

### LINKED ISSUE #255: Rework `total_earnings` to `scheduled_rewards`
We have recently reworked how rewards are paid to Station operators to use the model "push once a month". Station Desktop no longer shows "total earnings" but instead displays "scheduled rewards", see https://github.com/filecoin-station/desktop/pull/1082

We need to make similar changes in Station Core and possibly in the underlying modules (Zinnia, Bacalhau) too.

Keywords to search for in the codebase: `totalEarnings`, `total_earnings`.

Slack threads started by our users:
- https://filecoinproject.slack.com/archives/C03S6LXSRB8/p1699528485011289
- https://filecoinproject.slack.com/files/U02N6KCL19S/F0652MPRRBN/image.png
- https://filecoinproject.slack.com/archives/C04N76Y689M/p1699512083801239

juliangruber:
Closes #255


bajtos:
@juliangruber I'd like to discuss the high-level picture first.

(1)
It is possible to use the same wallet address for multiple Station Core instances. However, it's not possible to tell how many rewards were earned by which instance. All we have is the total sum for all Stations using this wallet address.

I expect this is going to create confusion for Station Core users.

For example, if I were running a bunch of Station Core instances, aggregating their metrics and rendering them in a dashboard, I would calculate my expected earnings as `sum(scheduledRewards)`. 

How can we clarify what the number in `scheduledRewards` means? Maybe we can rename it something else, e.g. `totalRewardsScheduledForAddress` (preferably find a more succinct option). 

Either way, it's IMO important to explain this in our docs (the README file?).

(2)
Should we leverage this new metrics produced by Station Core in Station Desktop? We can rework the way how Desktop obtains the "scheduled rewards" value to fetch it from Station Core state instead the RPC API.

The upside is that we avoid duplication.

The downside is a slightly more complex integration and closing the door to move RPC API calls querying on-chain state from `main` to `rendered`, which would simplify this part a lot.

juliangruber:
> How can we clarify what the number in scheduledRewards means? Maybe we can rename it something else, e.g. totalRewardsScheduledForAddress (preferably find a more succinct option).

Good call! I agree this would be clearer. What about `rewardsScheduledForAddress`?

> Should we leverage this new metrics produced by Station Core in Station Desktop? We can rework the way how Desktop obtains the "scheduled rewards" value to fetch it from Station Core state instead the RPC API.

I'm in favor of reading this from Station Core, so that we only have this logic in one place.

bajtos:
> `rewardsScheduledForAddress`

SGTM!

If we are not going to use this metric in Station Desktop, then does Station Core need to periodically emit it?

An alternative to consider: add a new CLI command to find out what's the current amount.

lib/zinnia.js
```diff
@@ -43,6 +46,22 @@ export async function run ({
...
+  const contract = new ethers.Contract(
+    '0xaaef78eaf86dcf34f275288752e892424dda9341',
+    JSON.parse(
+      await fs.readFile(
+        fileURLToPath(new URL('./abi.json', import.meta.url)),
+        'utf8'
+      )
+    ),
+    provider
+  )
```
bajtos, line 63: I find it weird that it's the Zinnia module that handles querying the smart contract for scheduled rewards.

In my mind, this functionality is not specific to Zinnia (or Bacalhau or any other module) and, therefore, it belongs to Station Core itself.

What are the arguments in favour of placing this new logic here?

juliangruber:
> If we are not going to use this metric in Station Desktop, then does Station Core need to periodically emit it?

Once `core` exposes it we can remove this exact logic from Desktop again, so that it only lives in one place. Wdyt?

lib/zinnia.js
juliangruber: This is a good question! My initial reasoning was that bacalhau doesn't use this, so it doesn't belong in `core`, above all modules. And who knows, maybe we will even add another trusted module in the future?

Otoh, it doesn't belong into Zinnia either, because it's the module's decision whether to use Meridian. `ethers` might even run in Zinnia, since for this we don't need to sign messages - or we could just use the jsonrpc endpoint directly - then we only need `fetch`. I think for this we need a way for Zinnia modules to log metrics.

juliangruber:
@bajtos hindsight please. I would like to get this out as a QoS improvement. I think the interface isn't going to change, only where the code lives - which we can address next week.

lib/zinnia.js
bajtos: Fair enough. Please update our checklist for deploying the new Meridian version to remind us that we need to update this place in Station Core, too?

lib/zinnia.js
```diff
@@ -91,42 +119,47 @@ export async function run ({
...
-              totalJobsCompleted: event.total,
-              totalEarnings: '0'
-            })
-            break
+        case 'jobs-completed':
+          onMetrics({
+            totalJobsCompleted: event.total,
+            rewardsScheduledForAddress:
+              (await contract.rewardsScheduledFor(FIL_WALLET_ADDRESS))
+                .toString()
```
bajtos, line 154: IIUC, if we cannot fetch the rewards, then the entire `handleEvents` call fails with an error. I think this is brittle: if `text` contains multiple events, the first one is `jobs-completed`, and the contract query fails, then we end up not processing all of the events.

Proposal:
- Extract a new function to query the scheduled rewards
- Handle RPC errors inside this function

bajtos (commented):
Let's improve error handling please 👇🏻

lib/zinnia.js
juliangruber: already part of https://github.com/filecoin-station/spark-api/blob/main/docs/impact-evaluator.md :)

lib/zinnia.js
juliangruber: Good call! Addressed in https://github.com/filecoin-station/core/pull/267

bajtos:
> > If we are not going to use this metric in Station Desktop, then does Station Core need to periodically emit it?
> 
> Once `core` exposes it we can remove this exact logic from Desktop again, so that it only lives in one place. Wdyt?

Initially, I as indifferent:

On one hand, removing the duplication is good.

On the other hand, there is also value in moving all on-chain querying into the Station Desktop front-end so that we don't have to maintain so much Electron IPC stuff.
- The backend (`main`) provides a IPC method for obtaining the public address of the built-in wallet
- The backend also provides addresses of Meridian smart contracts for querying the balances
- The front-end creates Ethers.js clients for these smart contracts and periodically queries the state

As I wrote this, I realised it may not be feasible to query "scheduled rewards" from the front-end if different modules have different smart contract ABIs and even different methods to query the balance. We may need to move this logic down into each Zinnia module, so that Station does not have any module-specific logic. In that case, it makes a lot of sense that Station Desktop should consume `rewardsScheduledForAddress` metric provided by Station Core.

So, let's do it!

→ https://github.com/filecoin-station/desktop/issues/1134

bajtos:
> Otoh, it doesn't belong into Zinnia either, because it's the module's decision whether to use Meridian. `ethers` might even run in Zinnia, since for this we don't need to sign messages - or we could just use the jsonrpc endpoint directly - then we only need `fetch`. I think for this we need a way for Zinnia modules to log metrics.

Yeah, I agree.

Opened a GH issue to implement this new feature in Zinnia: https://github.com/filecoin-station/zinnia/issues/421

bajtos:
> > Otoh, it doesn't belong into Zinnia either, because it's the module's decision whether to use Meridian. `ethers` might even run in Zinnia, since for this we don't need to sign messages - or we could just use the jsonrpc endpoint directly - then we only need `fetch`. I think for this we need a way for Zinnia modules to log metrics.
> 
> Yeah, I agree.
> 
> Opened a GH issue to implement this new feature in Zinnia: [filecoin-station/zinnia#421](https://github.com/filecoin-station/zinnia/issues/421)

Actually, this requires more work besides Zinnia, I opened a higher-level issue to keep track of all subtasks.

- https://github.com/filecoin-station/core/issues/275

PR #264, opened 2023-11-17
Fix retry wallet address check

juliangruber:
@bajtos hindsight please

This fixes an error in Station Desktop, where you have it launch at boot, but your network isn't ready yet -> `core` fails to launch and `desktop` doesn't recover.

Alternatives considered:

- Remove address check from `core` when it's launched by `desktop`, which will only supply Station-created addresses. This would be the simplest solution. Need to double check with legal if we may add this option. Downside: If one of the Station wallets gets sanctioned, this doesn't catch it.
- Add retry logic to Station Desktop. I think this would be nicer overall, but a more difficult change. Wdyt?

### DIFF
--- a/commands/station.js
+++ b/commands/station.js
@@ -32,8 +32,13 @@ export const station = async ({ json, experimental }) => {
   ) {
     panic('FIL_WALLET_ADDRESS must start with f410 or 0x')
   }
-  const fetchRes = await fetch(
-    `https://station-wallet-screening.fly.dev/${FIL_WALLET_ADDRESS}`
+  const fetchRes = await pRetry(
+    () => fetch(`https://station-wallet-screening.fly.dev/${FIL_WALLET_ADDRESS}`),
+    {
+      retries: 1000,
+      onFailedAttempt: () =>
+        console.error('Failed to check FIL_WALLET_ADDRESS address. Retrying...')
+    }
   )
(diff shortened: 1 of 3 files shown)


commands/station.js
```diff
@@ -32,8 +32,13 @@ export const station = async ({ json, experimental }) => {
...
     panic('FIL_WALLET_ADDRESS must start with f410 or 0x')
   }
-  const fetchRes = await fetch(
-    `https://station-wallet-screening.fly.dev/${FIL_WALLET_ADDRESS}`
+  const fetchRes = await pRetry(
+    () => fetch(`https://station-wallet-screening.fly.dev/${FIL_WALLET_ADDRESS}`),
+    {
+      retries: 1000,
+      onFailedAttempt: () =>
+        console.error('Failed to check FIL_WALLET_ADDRESS address. Retrying...')
```
bajtos, line 40: Nitpick, feel free to ignore:

IMO, "Failed to validate... " would be a more accurate message than "Failed to check..."

bajtos (commented):
LGTM.

If the computer is not online, then SPARK won't be able to connect to the orchestrator and the Station will not be serving any jobs. I think it's fine to delay the startup until we are online.

commands/station.js
juliangruber: https://github.com/filecoin-station/core/pull/268

PR #267, opened 2023-11-20
Fix scheduled rewards display

juliangruber:
There was an issue in the metrics merging logic (shipped too fast :|), it was always showing 0. Fixed this:

```
{
  "totalJobsCompleted": 680,
  "rewardsScheduledForAddress": "0.041033208757289921"
}
```

### DIFF
--- a/lib/metrics.js
+++ b/lib/metrics.js
@@ -51,6 +51,8 @@ export class Metrics {
     }
     for (const [, metrics] of this.moduleMetrics) {
       mergedMetrics.totalJobsCompleted += metrics.totalJobsCompleted
+      mergedMetrics.rewardsScheduledForAddress +=
+        metrics.rewardsScheduledForAddress
     }
     const isChanged = this.mergedMetrics === null ||
       Object
--- a/lib/metrics.js
+++ b/lib/metrics.js
@@ -6,7 +6,7 @@ export class MetricsEvent {
   /**
    * @param {Object} options
    * @param {Number} options.totalJobsCompleted
-   * @param {String} options.rewardsScheduledForAddress
+   * @param {bigint} options.rewardsScheduledForAddress
    */
   constructor ({ totalJobsCompleted, rewardsScheduledForAddress }) {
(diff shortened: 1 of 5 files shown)


lib/metrics.js
```diff
@@ -47,10 +47,12 @@ export class Metrics {
     this.moduleMetrics.set(moduleName, metrics)
     const mergedMetrics = {
       totalJobsCompleted: 0,
-      rewardsScheduledForAddress: '0'
+      rewardsScheduledForAddress: 0n
     }
     for (const [, metrics] of this.moduleMetrics) {
       mergedMetrics.totalJobsCompleted += metrics.totalJobsCompleted
+      mergedMetrics.rewardsScheduledForAddress +=
+        metrics.rewardsScheduledForAddress
```
bajtos, line 55: Composing rewards is tricky.
- If you run both Station Core and Station Desktop, this metric will show rewards for both apps combined together.
- If let's say Bacalhau starts paying rewards, I think it's safe to assume they will use a different smart contract, therefore the Bacalhau module will report different scheduled rewards than SPARK, and thus it's safe to sum these amounts as you are doing here.

I think it will be good to point out this complexity by adding a code comment.

WDYT?

bajtos (approved):

lib/metrics.js
juliangruber: > If you run both Station Core and Station Desktop, this metric will show rewards for both apps combined together.

This is true, but independent of merging module metrics here, right?

> If let's say Bacalhau starts paying rewards, I think it's safe to assume they will use a different smart contract, therefore the Bacalhau module will report different scheduled rewards than SPARK, and thus it's safe to sum these amounts as you are doing here.

I'll add a code comment 👍

bajtos: > > If you run both Station Core and Station Desktop, this metric will show rewards for both apps combined together.
> 
> This is true, but independent of merging module metrics here, right?

Yes, I think so.

PR #288, opened 2023-12-05
bacalhau: add arm support. closes #286

### LINKED ISSUE #286: Unable to install core in Raspberry Pi 3
I am trying to install code in Raspberry Pi 3 running Node version 18.1 and NPM 8.8.0.
I tried running npm install -g @filecoin-station/core but it failed with 👍🏾 
Screenshot 2023-12-04 at 20 24 46

juliangruber:
Closes #286

### DIFF
--- a/lib/bacalhau.js
+++ b/lib/bacalhau.js
@@ -21,7 +21,8 @@ export async function install () {
       { platform: 'darwin', arch: 'arm64', asset: `bacalhau_${DIST_TAG}_darwin_arm64.tar.gz` },
       { platform: 'linux', arch: 'x64', asset: `bacalhau_${DIST_TAG}_linux_amd64.tar.gz` },
       { platform: 'linux', arch: 'arm64', asset: `bacalhau_${DIST_TAG}_linux_arm64.tar.gz` },
-      { platform: 'win32', arch: 'x64', asset: `bacalhau_${DIST_TAG}_windows_amd64.tar.gz` }
+      { platform: 'win32', arch: 'x64', asset: `bacalhau_${DIST_TAG}_windows_amd64.tar.gz` },
+      { platform: 'linux', arch: 'arm', asset: `bacalhau_${DIST_TAG}_linux_arm6.tar.gz` }
     ]
   })
 }


lib/bacalhau.js
```diff
@@ -21,7 +21,8 @@ export async function install () {
       { platform: 'darwin', arch: 'arm64', asset: `bacalhau_${DIST_TAG}_darwin_arm64.tar.gz` },
       { platform: 'linux', arch: 'x64', asset: `bacalhau_${DIST_TAG}_linux_amd64.tar.gz` },
       { platform: 'linux', arch: 'arm64', asset: `bacalhau_${DIST_TAG}_linux_arm64.tar.gz` },
-      { platform: 'win32', arch: 'x64', asset: `bacalhau_${DIST_TAG}_windows_amd64.tar.gz` }
+      { platform: 'win32', arch: 'x64', asset: `bacalhau_${DIST_TAG}_windows_amd64.tar.gz` },
+      { platform: 'linux', arch: 'arm', asset: `bacalhau_${DIST_TAG}_linux_arm6.tar.gz` }
```
bajtos, line 25: Would you mind keeping this list sorted by platform? (Add the new line before the win32 entry, not after.)

```suggestion
      { platform: 'linux', arch: 'arm', asset: `bacalhau_${DIST_TAG}_linux_arm6.tar.gz` },
      { platform: 'win32', arch: 'x64', asset: `bacalhau_${DIST_TAG}_windows_amd64.tar.gz` }
```

bajtos, line 25: I see that Bacalhau supports both Arm6 and Arm7. Should we support both?

FWIW, in [this discussion](https://forum.openframeworks.cc/t/raspberry-pi-3-arm6-or-arm7/26800) from 2017, they say
> On the Raspberry 2 although it’s architecture is arm7 raspbian only supports arm6 by now so be sure to download that version as is explained in the getting started guide. If you want to use arm7 binaries with the raspberry PI 2 you can try other distributions like archlinux or ubuntu core and the arm7 download of OF but this is still not very well tested so some changes might be required to the original install scripts.

I also found this official blog post from 2022 which provides detailed explanations about different ARM versions in different Raspberry Pi versions:
https://www.raspberrypi.com/news/raspberry-pi-os-64-bit/

My understanding is that rPi users have two options:
- Run an older version of the OS in arm6 mode (32bit)
- Run a newer version of the OS in arm64 mode (64bit)

rPi models starting from rPi3 and rPi Zero 2 support arm64.

In that light:
- I am questioning whether we need to support 32bit ARM at all.
- If we want to support 32bit ARM, then `arm6` is the option that gives us the best compatibility.

bajtos (changes requested):
This solves the installation of Bacalhau on arm6, but how about Zinnia?

Zinnia doesn't provide binaries for arm6. Building such binaries will be very expensive because AFAIK rusty_v8 does not provide pre-built binaries for that platform, and therefore, we would have to build v8 from sources.

juliangruber:
Ah, I wasn't aware that Zinnia doesn't support arm. Then this is here doesn't add any value

juliangruber:
https://github.com/filecoin-station/core/issues/286#issuecomment-1841468658

PR #305, opened 2024-01-02
Add auth to glif requests

juliangruber:
Ref https://filecoinproject.slack.com/archives/C023K7D9GAX/p1703679868805479

### DIFF
--- a/lib/zinnia.js
+++ b/lib/zinnia.js
@@ -57,8 +57,15 @@ export async function run ({
   onActivity,
   onMetrics
 }) {
+  const fetchRequest = new ethers.FetchRequest(
+    'https://api.node.glif.io/rpc/v1'
+  )
+  fetchRequest.setHeader(
+    'Authorization',
+    'Bearer P1Kqr5vHZWNEISNyzfIqx8FbBwdGzaTG5Fn3pzkpHCU='
+  )
   const provider = new ethers.JsonRpcProvider(
-    'https://api.node.glif.io/rpc/v1',
+    fetchRequest,
     null,
     { batchMaxCount: 1 }
   )


bajtos (approved):
Hmm, what are the consequences of making the access token public? What happens if somebody takes our token and uses it elsewhere, thus exhausting our rate limit?

Can we change Station Core to generate a new access token when it starts?

Alternatively, can we request Core operators to provide the token in a new ENV var? (Station Desktop can handle that transparently for desktop users.)

I am okay with landing this fix as a stop-gap solution, but I would like us to consider the ramifications.

juliangruber:
To generate a token, you need to sign up for https://auth.node.glif.io/. I think this can be automated, since the site just uses your wallet to sign messages, and I assume then submits them to an API. There is no public API however, so I doubt this is a safe way to go.

There is no rate limiting yet, Glif is rolling out this change to understand usage patterns. I agree that leaking the token is a risk - similar to how we have to leak our InfluxDB tokens.

We could request people to provide their own token, yeah. For Core it could work, but again with Desktop we face this issue that there's no automated way to generate it. Since you can't connect your Station wallet with Metamask, people will need to go to that website and bring their own wallet.

Therefore, I currently don't see an alternative. Wdyt?

juliangruber:
I'm going to land this now since you gave an approval. We can always rotate the token later. Happy to continue discussing in this thread.

juliangruber:
https://github.com/filecoin-station/core/releases/tag/v16.3.2

bajtos:
I am proposing to revisit this topic in a few week's time. Hopefully, there will be more information and/or API docs from Glif by that time.

PR #323, opened 2024-01-29
Add `voyager` module

juliangruber:


lib/zinnia.js
```diff
@@ -171,14 +184,16 @@ export async function run ({
             type: 'info',
             message: 'Updated Zinnia module source files, restarting...'
           })
-          childProcess.kill()
+          for (const childProcess of childProcesses) {
+            childProcess.kill()
+          }
```
juliangruber, line 189: For simplicity, always tear down all processes, instead of only recreating the one that needs to quit

DiegoRBaquero (approved):

lib/zinnia.js
```diff
@@ -48,6 +54,7 @@ const updateAllSourceFiles = async () => {
   const modules = await Promise.all(
     Object
       .values(ZINNIA_MODULES)
+      .filter(({ module }) => module.includes(MODULE_FILTER))
```
bajtos, line 57: I would expect `MODULE_FILTER` to support more than one module, e.g. `MODULE_FILTER=spark,bacalhau`.

I would also expect this filter to be applied to non-Zinnia modules like Bacalhau.

At the same time, I think this is all relatively low priority. I propose to either
- Change the env var name to better convey the intent (e.g. `SINGLE_ZINNA_MODULE`)
- Drop module filtering completely and wait until there is user demand

bajtos, line 57: I find `include()` check too hacky and possibly confusing. One can specify `MODULE_FILTER=r` to run both Zinnia modules (voyageR, spaRk).

Why not to check for equality? `MODULE_FILTER ? module === MODULE_FILTER : true`

bajtos (changes requested):

lib/zinnia.js
juliangruber: I'll change it to strict equality 👍

juliangruber: It just makes the implementation more complicated, because we need to check for the empty filter case too

bajtos: > If you work on the spark or voyager integration, I'm sure you'll find use for it too.

I wasn't aware this is something you found useful yourself. I agree to keep this feature.

> As we only have two modules, filtering multiple via `a,b` isn't required. I was planning to add this once we have a 3rd module. To me nothing about `MODULE_FILTER` communicates that I can pass a comma separated list. Maybe we can rename it to `MODULE_GREP`?

I see. I would expect MODULE_GREP to support regular expressions. MODULE_FILTER is better in this aspect.

My main concern is about making this feature easy to understand and avoid confusion.

I don't have any more suggestions. The current implementation seems to be the best we can think of now. 🤷🏻 

> The Bacalhau module is an experimental one and not being updated (and not being used). I'm rather leaning towards removing it instead.

+1 to remove it so that we can simplify our work by not having to consider Bacalhau for consistency across all modules.

bajtos: > It just makes the implementation more complicated, because we need to check for the empty filter case too

Yeah. Maybe you can extract a small helper function to keep this more DRY?

```js
.filter(({ module }) => matchesModuleFilter(MODULE_FILTER, module))
```

README.md
```diff
@@ -81,6 +81,10 @@ the configuration options described in
   `0x000000000000000000000000000000000000dEaD`. Please note that any earnings
   sent there will be lost.
 
+- `MODULE_FILTER` _(string; optional)_: Filter Zinnia modules by name. Eg:
+  - `spark`
+  - `voyager`
```
bajtos, line 86:
```suggestion
- `MODULE_FILTER` _(string; optional)_: Run only the Zinnia module with the given name. Eg:
  - `MODULE_FILTER=spark`
  - `MODULE_FILTER=voyager`
```

lib/zinnia.js
```diff
@@ -189,10 +205,15 @@ export async function run ({
       }
     })(),
     (async () => {
-      const [code] = await once(childProcess, 'close')
+      const [code] = await Promise.race(
```
bajtos, line 208: I think we should wait for all processes to close before we continue, otherwise we can end up with a module having two processes - the old process still writing the output, the new process already starting.

```suggestion
      const [code] = await Promise.allSettled(
```

lib/zinnia.js
```diff
@@ -121,36 +126,45 @@ export async function run ({
...
+      }).catch(err => {
+        console.error(err)
+        Sentry.captureException(err)
+      })
+    })
+    childProcess.stderr.pipe(process.stderr, { end: false })
+
+    childProcess.on('exit', (code, signal) => {
+      exitReason = signal ? `via signal ${signal}` : `with code: ${code}`
+      const msg = `Zinnia exited ${exitReason}`
```
bajtos, line 164: There will be multiple Zinnia processes, one per module. This message will be difficult to troubleshoot because we don't know which module exited with the reported reason.

I think we should change all messages and rename `Zinnia` to the module name (`Spark` or `Voyager`.)

lib/zinnia.js
```diff
@@ -121,36 +126,45 @@ export async function run ({
...
-  const childProcess = execa(zinniadExe, modules, {
-    cwd: moduleBinaries,
-    env: {
-      FIL_WALLET_ADDRESS,
-      STATE_ROOT,
-      CACHE_ROOT
-    }
-  })
+  const childProcesses = []
+  let exitReason
```
bajtos, line 130: We need to track the exit reason for each process.

lib/zinnia.js
```diff
@@ -171,14 +185,16 @@ export async function run ({
...
+            childProcess.kill()
+          }
           return
         }
       }
     })(),
     (async () => {
       try {
-        await childProcess
+        await Promise.race(childProcesses)
```
bajtos, line 197: IIUC, this will return as soon as the first process exits, and the other promises will stay running in the background. What is going to happen with those promises? In particular, if the promise slower to finish ends up being rejected, how will we handle that error?

Looking from a higher-level perspective, we need to decide how to handle the following cases:

- Both (all) modules start and happily run
- One module starts, but the other crashes
- Both (all) modules crashes

The answers are not clear to me from reading the code.

bajtos (changes requested):

lib/zinnia.js
juliangruber: > My main concern is about making this feature easy to understand and avoid confusion.
> I don't have any more suggestions. The current implementation seems to be the best we can think of now. 🤷🏻

Ok let's ship it like this. We can always iterate on it. I don't want to block the Voyager integration on this simple env var.

> > The Bacalhau module is an experimental one and not being updated (and not being used). I'm rather leaning towards removing it instead.
> 
> +1 to remove it so that we can simplify our work by not having to consider Bacalhau for consistency across all modules.

Cool! https://github.com/filecoin-station/core/pull/326

juliangruber: done https://github.com/filecoin-station/core/pull/323/commits/9c9aee6ea34df95a93e49b286b3faf8f7410cdb5 🙏

lib/zinnia.js
juliangruber: Good idea. Changed it to `Promise.all()` (don't see the need for `allSettled`, correct me if I'm wrong) and ensured the on exit listener above kills all the procs when one exits

lib/zinnia.js
juliangruber: Good idea! Done in https://github.com/filecoin-station/core/pull/323/commits/22b89c79a1e7d8f802dce7db385c59bfa748e0d8 (forgive the busy commit)

lib/zinnia.js
juliangruber: Fixed in https://github.com/filecoin-station/core/pull/323/commits/22b89c79a1e7d8f802dce7db385c59bfa748e0d8, `exitReason`'s scope was wrong

lib/zinnia.js
juliangruber: See the current version of the code (I added some changes after your feedback). When one process exits, all others will be killed too.

In the previous version of the code, once the first child process closed stdout, all others were killed.

This means that if one Zinnia process is unhealthy, all will be taken down. I think we can iterate this and decouple the processes completely, if we see that this is causing problems.

lib/zinnia.js
```diff
@@ -196,28 +200,29 @@ export async function run ({
...
       } catch (err) {
-        if (shouldRestart) return
-        const errorMsg = err instanceof Error ? err.message : '' + err
-        const message = `Cannot start Zinnia: ${errorMsg}`
-        onActivity({ type: 'error', message })
-        maybeReportCrashToSentry(err)
-        throw err
+        if (!shouldRestart) {
+          const errorMsg = err instanceof Error ? err.message : '' + err
+          const message = `Cannot start Zinnia: ${errorMsg}`
```
bajtos, line 206:
```suggestion
          const message = `Cannot start ${capitalize(module)}: ${errorMsg}`
```

lib/zinnia.js
bajtos: > When one process exits, all others will be killed too.
> This means that if one Zinnia process is unhealthy, all will be taken down. I think we can iterate this and decouple the processes completely, if we see that this is causing problems.

SGTM 👍🏻 

Let's ship a simple version and then iterate based on observing real-world behaviour.

bajtos (approved):
👏🏻

lib/zinnia.js
juliangruber: there's no `module` available here unfortunately, since this is past the modules loop. We only have an array of child processes, and in the `catch` handler can't tell which exited (afaik)

juliangruber: happy to follow up :)

bajtos: I see. Let's see if this ever becomes a problem in the real world.

PR #324, opened 2024-01-30
Fix redundant module updates

juliangruber:
By storing the latest installed module version in a text file, prevent redundant updates.

### DIFF
--- a/lib/modules.js
+++ b/lib/modules.js
@@ -115,9 +115,9 @@ async function getLatestDistTag (repo) {
   return body.tag_name
 }
 
-async function getLastSeenModuleDistTag ({ module, latestModuleVersions }) {
+async function getLastSeenModuleDistTag ({ module, moduleVersionsDir }) {
   try {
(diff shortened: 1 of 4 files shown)


lib/modules.js
```diff
@@ -115,19 +115,34 @@ async function getLatestDistTag (repo) {
   return body.tag_name
 }
 
-const lastModuleDistTag = new Map()
+async function getLastSeenModuleDistTag ({ module, latestModuleVersions }) {
```
bajtos, line 118: I find it a bit confusing that `latestModuleVersions` is a directory path. When I see this name, I expect a collection of version numbers.

How about changing this name to something like `latestVersionsDir` or `moduleVersionsDir`? `latestModuleVersionsDir` is an option too, although a bit long.

bajtos (approved):
Nice!

lib/modules.js
juliangruber: I found this confusing as well, thanks will change!

bajtos (approved):
:shipit:

PR #326, opened 2024-01-31
Remove Bacalhau

juliangruber:
The module never left `experimental` status, and now is causing some inconsistency issues in the codebase.

Ref https://github.com/filecoin-station/core/pull/323#discussion_r1472596922


commands/station.js
```diff
@@ -103,14 +101,7 @@ export const station = async ({ json, experimental }) => {
...
   if (experimental) {
-    modules.push(pRetry(() => bacalhau.run({
-      FIL_WALLET_ADDRESS,
-      storagePath: join(paths.moduleCache, 'bacalhau'),
-      onActivity: activity => {
-        activities.submit({ source: 'Bacalhau', ...activity })
-      },
-      onMetrics: m => metrics.submit('bacalhau', m)
-    }), { retries: 1000 }))
+    // No experimental module available at this point
```
bajtos, line 104: Should we print a message to let the user know their CLI arg is being ignored?

bajtos (approved):
LGTM.

What do you think about our "core on fly" deployment? Should we stop the VM running Station Core with experimental modules enabled? Update Dependabot config to stop updating the Core docker image version used by that VM?

juliangruber:
> LGTM.
> 
> What do you think about our "core on fly" deployment? Should we stop the VM running Station Core with experimental modules enabled? Update Dependabot config to stop updating the Core docker image version used by that VM?

Good call, I propose we remove the experimental app (can always revert later)

commands/station.js
juliangruber: great detail!

PR #328, opened 2024-02-02
Download updates via ipfs, not github

juliangruber:
- [x] Download updates via ipfs
- [x] Verify download

### DIFF
--- a/lib/modules.js
+++ b/lib/modules.js
@@ -167,23 +167,17 @@ export async function updateSourceFiles ({ module, ipnsKey, moduleVersionsDir })
         return block.bytes
       }
     })
-    let entryCount = 0
-    for await (const entry of entries) {
-      if (++entryCount > 1) {
-        console.error('Warning: Unexpected multiple entries in CAR archive')
-        console.error('Skipping...')
-        break
-      }
-      // Depending on size, entries might be packaged as `file` or `raw`
-      // https://github.com/web3-storage/w3up/blob/e8bffe2ee0d3a59a977d2c4b7efe425699424e19/packages/upload-client/src/unixfs.js#L11
-      if (entry.type === 'file' || entry.type === 'raw') {
-        // `{ strip: 1}` tells tar to remove the top-level directory (e.g. `mod-peer-checker-v1.0.0`)
-        await pipeline(
-          entry.content(),
-          gunzip(),
-          tar.extract(outDir, { strip: 1 })
-        )
-      }
+    const { value: entry } = await entries.next()
+    assert(entry, `No entries in ${module} archive`)
+    // Depending on size, entries might be packaged as `file` or `raw`
+    // https://github.com/web3-storage/w3up/blob/e8bffe2ee0d3a59a977d2c4b7efe425699424e19/packages/upload-client/src/unixfs.js#L11
+    if (entry.type === 'file' || entry.type === 'raw') {
+      // `{ strip: 1}` tells tar to remove the top-level directory (e.g. `mod-peer-checker-v1.0.0`)
+      await pipeline(
+        entry.content(),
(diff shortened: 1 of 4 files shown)


lib/modules.js
```diff
@@ -126,59 +113,78 @@ async function getLastSeenModuleDistTag ({ module, moduleVersionsDir }) {
...
-    await pipeline(res.body, gunzip(), tar.extract(outDir, { strip: 1 }))
+    const reader = await CarReader.fromIterable(res.body)
+    const entries = exporter(cid, {
+      async get (blockCid) {
+        const block = await reader.get(blockCid)
+        try {
+          await validateBlock(block)
+        } catch (err) {
+          throw new Error(
+            `Invalid block ${blockCid} of root ${cid}. ${err.stack}`
```
juliangruber, line 164: The typescript setup disallows passing a second argument to `Error()`

lib/modules.js
```diff
@@ -126,59 +113,78 @@ async function getLastSeenModuleDistTag ({ module, moduleVersionsDir }) {
...
+      // `{ strip: 1}` tells tar to remove the top-level directory (e.g. `mod-peer-checker-v1.0.0`)
+      await pipeline(
+        entry.content(),
+        gunzip(),
+        tar.extract(outDir, { strip: 1 })
+      )
+    }
   } catch (err) {
-    await rmdir(outDir, { recursive: true })
+    await rm(outDir, { recursive: true })
```
juliangruber, line 183: node threw a deprecation warning and suggested to move to `rm` instead

juliangruber:
@bajtos hindsight please. I'm going ahead as the code changes are fairly straight forward, and this fixes a current production issue.

lib/modules.js
bajtos: I think this is caused by our TypeScript configuration targeting ES2021 - I guess the Error constructor did not have the second argument back then.

https://github.com/filecoin-station/core/blob/1e9bb5e1dc4fa46036f34d66aa56b386a5c03ff9/tsconfig.json#L6-L7

Depending on what's our minimum supported Node.js version, we can bump this to a newer ES standard.

lib/modules.js
```diff
@@ -126,59 +113,78 @@ async function getLastSeenModuleDistTag ({ module, moduleVersionsDir }) {
...
+    const { value: entry } = await entries.next()
+    assert(entry, `No entries in ${module} archive`)
+    // Depending on size, entries might be packaged as `file` or `raw`
+    // https://github.com/web3-storage/w3up/blob/e8bffe2ee0d3a59a977d2c4b7efe425699424e19/packages/upload-client/src/unixfs.js#L11
+    if (entry.type === 'file' || entry.type === 'raw') {
+      // `{ strip: 1}` tells tar to remove the top-level directory (e.g. `mod-peer-checker-v1.0.0`)
+      await pipeline(
+        entry.content(),
+        gunzip(),
+        tar.extract(outDir, { strip: 1 })
```
bajtos, line 179: What happens when `ourDir` already exists and the new module version no longer contains some files? (E.g. we delete a source code file in Spark and then publish a new version.)

Are we going to end up with unused old source code files?

lib/modules.js
juliangruber: According to https://stackoverflow.com/a/67371788, we can't upgrade as long as we're using node 16. But! Node 16 is past EOL so let's update the minimum engine

juliangruber: https://github.com/filecoin-station/core/pull/335

lib/modules.js
juliangruber: good point! I think this can happen. https://github.com/filecoin-station/core/issues/336

PR #329, opened 2024-02-02
fix store module sources in CACHE_ROOT

juliangruber:
The main directory isn't always writeable, eg when mounted inside an electron app. Therefore, store files inside CACHE_ROOT

### DIFF
--- a/lib/modules.js
+++ b/lib/modules.js
@@ -1,7 +1,7 @@
 import os from 'node:os'
 import assert from 'node:assert'
 import { join } from 'node:path'
-import { mkdir, chmod, rm, readFile, writeFile } from 'node:fs/promises'
+import { mkdir, chmod, rm, readFile, writeFile, stat } from 'node:fs/promises'
 import { fetch } from 'undici'
 import { pipeline } from 'node:stream/promises'
 import gunzip from 'gunzip-maybe'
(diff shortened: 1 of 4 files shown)


lib/modules.js
```diff
@@ -131,8 +136,13 @@ export async function updateSourceFiles ({ module, ipnsKey, moduleVersionsDir })
   const cid = await getLatestCID(ipnsKey)
   const isUpdate = lastSeenCID !== cid
   if (!isUpdate) {
-    console.error(`[${module}]  ✓ no update available`)
-    return isUpdate
+    try {
+      await stat(outDir)
```
juliangruber, line 140: This is another precaution: Station Desktop was also failing because according to the `latest` file we had an installation of the module, but the files weren't on disk.

juliangruber:
@bajtos hindsight please

bajtos (commented):
In hindsight, it makes perfect sense that we must write module sources to a different directory 🤦🏻 

I am not sure which directory (CACHE_DIR or STATE_DIR) is the right one, but I also think this does not matter much.

The changes LGTM 👍🏻

juliangruber:
> I am not sure which directory (CACHE_DIR or STATE_DIR) is the right one, but I also think this does not matter much.

I was also wondering about this. To me it feels more like caching the source files on disk, than treating them as app state (since when the files are deleted, they can just be downloaded again and everything is as before)

PR #334, opened 2024-02-05
rotate telemetry token

juliangruber:
For now I've decided to split tokens by Core and Desktop, so that we can rotate them individually, and also name them by the core/desktop version they've been introduced in. We can iterate on this in the future.

@bajtos hindsight please 🙏


juliangruber:
https://github.com/filecoin-station/core/releases/tag/v18.1.3

PR #339, opened 2024-02-06
Fix `updateAllSourceFiles` signature

juliangruber:
@bajtos do you know why TS didn't fail on this?

Discovered via Sentry: https://protocol-labs-ip.sentry.io/issues/4949926341/?project=4504792315199488&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=5

### DIFF
--- a/lib/zinnia.js
+++ b/lib/zinnia.js
@@ -201,7 +201,10 @@ export async function run ({
       while (true) {
         await timers.setTimeout(10 * 60 * 1000) // 10 minutes
         try {
-          shouldRestart = await updateAllSourceFiles(moduleVersionsDir)
+          shouldRestart = await updateAllSourceFiles({
+            moduleVersionsDir,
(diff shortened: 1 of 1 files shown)


bajtos:
> @bajtos do you know why TS didn't fail on this?

The variable `moduleVersionsDir` has type `any`, therefore it's assignable to all types, including `{ moduleVersionsDir: string , moduleSourcesDir: string }`

Screenshot 2024-02-06 at 14 33 16

To catch mistakes like this, we can enable the TS option `noImplicitAny` ([docs](https://www.typescriptlang.org/tsconfig#noImplicitAny)). However, that would mean we can no longer use `any` in places where we feel types are not bringing enough value to justify the cost of writing JSDoc comments.

WDYT?

juliangruber:
Got it! And good question. I'll have to think about it

PR #343, opened 2024-02-07
Add gateway fallback

juliangruber:


juliangruber:
@bajtos hindsight please

lib/modules.js
```diff
@@ -145,21 +151,33 @@ export async function updateSourceFiles ({
...
-    )
+      if (!res.body) {
+        throw new Error(
+          `[${module}] Cannot fetch ${module} archive for ${cid}: no response body`
+        )
+      }
+      break
+    } catch (err) {
+      if (gateway === gateways[gateways.length - 1]) {
+        throw err
```
bajtos, line 176: I would throw a different error explaining that we have exhausted all fallback options to avoid potential confusion.

bajtos (commented):
LGTM 👍🏻

lib/modules.js
juliangruber: 👍  https://github.com/filecoin-station/core/pull/346

PR #346, opened 2024-02-08
Improve gateway error message

juliangruber:
Ref https://github.com/filecoin-station/core/pull/343#discussion_r1481701330

### DIFF
--- a/lib/modules.js
+++ b/lib/modules.js
@@ -173,7 +173,9 @@ export async function updateSourceFiles ({
       break
     } catch (err) {
       if (gateway === gateways[gateways.length - 1]) {
-        throw err
+        throw new Error(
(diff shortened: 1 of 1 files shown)


lib/modules.js
```diff
@@ -173,7 +173,9 @@ export async function updateSourceFiles ({
       break
     } catch (err) {
       if (gateway === gateways[gateways.length - 1]) {
-        throw err
+        throw new Error(
+          `[${module}] Can't download module sources from any gateway. Last error: ${err.stack}`
+        )
```
bajtos, line 178: Can we/should we store the original error in the "cause" property of the new error?

bajtos (approved):

lib/modules.js
juliangruber: good idea, we can use that now :)

PR #356, opened 2024-02-20
Add update contract addresses

juliangruber:
This makes core refresh the contracts used while it's running.

Following the 3 commits should be the best way to review.

### DIFF
--- a/lib/zinnia.js
+++ b/lib/zinnia.js
@@ -271,8 +271,8 @@ export async function run ({
           throw err
         }
         const newContracts = await getContractsWithRetry(provider, abi)
-        while (contracts.length > 0) contracts.pop()
-        for (const contract of newContracts) contracts.push(contract)
+        contracts.splice(0)
+        contracts.push(...newContracts)
       }
     })(),
     (async () => {
--- a/lib/zinnia.js
+++ b/lib/zinnia.js
@@ -102,6 +102,28 @@ async function getContractAddresses () {
   return revision.value.split('\n').filter(Boolean)
 }
 
+async function getContractsWithRetry (provider, abi) {
+  const contractAddresses = await pRetry(getContractAddresses, {
+    retries: 10,
+    onFailedAttempt: err => {
+      console.error(err)
+      console.error('Failed to get contract addresses. Retrying...')
+      if (String(err).includes('You are being rate limited')) {
+        const delaySeconds = 30 + (Math.random() * 60)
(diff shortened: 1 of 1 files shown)


lib/zinnia.js
```diff
@@ -251,6 +260,21 @@ export async function run ({
...
+        const jitter = Math.random() * 20_000 - 10_000 // +- 10 seconds
+        try {
+          await timers.setTimeout(delay + jitter, null, { signal })
+        } catch (err) {
+          if (err.name === 'AbortError') return
+          throw err
+        }
+        const newContracts = await getContractsWithRetry(provider, abi)
+        while (contracts.length > 0) contracts.pop()
+        for (const contract of newContracts) contracts.push(contract)
```
bajtos, line 275: Nitpick: this can be simplified.

```suggestion
        contracts.splice(0)
        contracts.push(...newContracts)
```

bajtos: Or even like this:

```suggestion
        contracts.splice(0, Infinity, ...newContracts)
```

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

juliangruber: Great suggestion! I committed the first one after seeing it, I think the next one is cool but not worth the extra effort now

juliangruber:
@bajtos is this a +1?

lib/zinnia.js
```diff
@@ -221,11 +221,20 @@ export async function run ({
   }
 
   let shouldRestart
+  const controller = new AbortController()
+  const { signal } = controller
 
   await Promise.all([
```
bajtos, line 227: The list of promises we are running in parallel is getting rather long. I find it difficult to understand the big picture because there are too many low-level implementation details obscuring the vision.

It would be great to refactor this code, extract smaller functions and simplify the `Promise.all` statement into something like this:

```js
await Promise.all([
  runModuleUpdater(),
  runContractAddressUpdater(),
  runChildProcesses(),
  waitForChildrenToClose(),
])
```

Such refactoring is out of scope of this PR, though.

bajtos (approved):
👍🏻

lib/zinnia.js
juliangruber: fantastic idea. I'll follow up!

juliangruber: https://github.com/filecoin-station/core/pull/358

PR #358, opened 2024-02-21
zinnia: refactor main loop

juliangruber:
Ref: https://github.com/filecoin-station/core/pull/356#discussion_r1496032902

### DIFF
--- a/lib/zinnia.js
+++ b/lib/zinnia.js
@@ -124,6 +124,98 @@ async function getContractsWithRetry (provider, abi) {
   })
 }
 
+const runUpdateSourceFilesLoop = async ({
+  signal,
+  onActivity,
+  childProcesses,
+  moduleVersionsDir,
+  moduleSourcesDir,
+  shouldRestart
+}) => {
+  while (true) {
+    const delay = 10 * 60 * 1000 // 10 minutes
+    const jitter = Math.random() * 20_000 - 10_000 // +- 10 seconds
+    try {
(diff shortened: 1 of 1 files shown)


bajtos (approved):
Awesome 👏🏻 

I did not review the details; I assume you just moved the source code into the new functions.

lib/zinnia.js
```diff
@@ -220,91 +312,33 @@ export async function run ({
...
-  let shouldRestart
+  const shouldRestart = {
+    value: false,
+    set (value) {
+      this.value = value
+    },
+    get () {
+      return this.value
+    }
+  }
```
juliangruber, line 323: @bajtos this is new: I turned `shouldRestart` into an object, so that functions that get a reference to it can still change its value.

juliangruber (commented):
@bajtos I've highlighted a place for review 🙏  Otherwise yes this is a straight port

lib/zinnia.js
bajtos: Yeah, I noticed. I have mixed feelings about this structure, but I also cannot come up with an option that would be clearly better. Let's keep your version for now. We can relatively easily refactor this place if we discover a better approach.

bajtos (approved):
:shipit:

lib/zinnia.js
juliangruber: I share your feelings. Sounds good!

PR #373, opened 2024-03-15
zinnia: fix end update loops outside of signal event

juliangruber:
I noticed looking at the `core-fly` logs that multiple update loops were running in parallel. This fixes the case when the abort controller is triggered but the async function that's currently running doesn't listen for it


bajtos (commented):
👍🏻

PR #377, opened 2024-03-19
fix: report more data to Sentry when Zinnia exits

bajtos:
Example issue in Sentry: https://protocol-labs-ip.sentry.io/issues/5058736376/?notification_uuid=f3501a23-8e05-490f-968c-d32f290de42e&project=4504792315199488&referrer=weekly_report


juliangruber (approved):
Let's give it a try 🥳

PR #389, opened 2024-04-02
Fix main loop not ending properly

juliangruber:
There are some issues with the zinnia loop end logic. See for example these logs from `core-fly`:

```
Mar 26 15:03:07 core-fly [spark]  ⇣ checking for updates
Mar 26 15:03:08 core-fly [voyager]  ⇣ checking for updates
Mar 26 15:03:08 core-fly [spark]  ⇣ checking for updates
Mar 26 15:03:08 core-fly [voyager]  ⇣ checking for updates
Mar 26 15:03:08 core-fly [voyager]  ✓ no update available
Mar 26 15:03:08 core-fly [spark]  ✓ no update available
Mar 26 15:03:08 core-fly [spark]  ✓ no update available
Mar 26 15:03:08 core-fly [voyager]  ✓ no update available
```

We can see that 2 update loops were running at once, which means that in a previous `run()` invocation the loop should have been aborted, but wasn't.

This PR fixes this by only using one method: `controller.abort()` for aborting, and `controller.signal.aborted` for checking if it was aborted. Previously, `shouldRestart.get()` was used in conjunction with the AbortController, and in some places only one was checked or set.

Should fix issues like https://filecoinproject.slack.com/archives/C03S6LXSRB8/p1711639505087089. This has also been observed on `core-fly`:

```
Mar 26 15:13:09 core-fly [3/26/2024, 2:13:09 PM] INFO  Updated Zinnia module source code, restarting...
```

(no logs following)

### DIFF
--- a/lib/zinnia.js
+++ b/lib/zinnia.js
@@ -139,12 +139,12 @@ async function getContractsWithRetry (provider, abi) {
 }
 
 const runUpdateSourceFilesLoop = async ({
+  controller,
   signal,
   onActivity,
   childProcesses,
   moduleVersionsDir,
-  moduleSourcesDir,
-  shouldRestart
+  moduleSourcesDir
 }) => {
   while (true) {
     const delay = 10 * 60 * 1000 // 10 minutes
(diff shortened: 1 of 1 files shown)


lib/zinnia.js
```diff
@@ -356,47 +356,36 @@ export async function run ({
...
-      shouldRestart
+      moduleSourcesDir
     }),
     runUpdateContractsLoop({ signal, provider, abi, contracts }),
-    catchChildProcessExit({ childProcesses, shouldRestart, onActivity }),
+    catchChildProcessExit({ childProcesses, onActivity, signal }),
     waitForStdioClose({ childProcesses, controller })
   ])
 
-  if (shouldRestart.get()) {
```
juliangruber, line 385: I could remove this condition because here we know we always want to restart

PR #391, opened 2024-04-02
Fix more zinnia loop issues

juliangruber:
The previous patch wasn't sufficient, core-fly got stuck after this:

Screenshot 2024-04-02 at 16 49 16

When we threw an error, no clean up happened. This is now fixed. And also, we now let execa kill all child processes through the AbortSignal, which helps tighten up the clean up further.

### DIFF
--- a/lib/zinnia.js
+++ b/lib/zinnia.js
@@ -146,7 +146,6 @@ const runUpdateSourceFilesLoop = async ({
   controller,
   signal,
   onActivity,
-  childProcesses,
   moduleVersionsDir,
   moduleSourcesDir
 }) => {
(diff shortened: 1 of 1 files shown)


bajtos (approved):
I am a bit concerned about the fragility of this orchestration layer. Is it perhaps time to improve `zinniad` to be finally able to run multiple modules, as it was meant to from the beginning?

Other than that, I don't see any obvious problems in these changes.

It would be great to document how to (manually?) test this orchestration logic so that we don't accidentally introduce regressions in the future.

juliangruber:
The hard thing is that I haven't been able to reproduce any of these issues. They happen after a while, on Fly.io.

I agree that this logic is showing to be a bit brittle, and that a more fundamental change is needed. Adding multi module support to zinniad will help for sure! There's also complexity in updating the source files, but with just one zinnia process it's less likely to get stuck.

PR #399, opened 2024-04-05
zinnia: fix update rewards loop order

juliangruber:
hindsight please, releasing asap as to not show rewards: 0

### DIFF
--- a/lib/zinnia.js
+++ b/lib/zinnia.js
@@ -211,18 +211,12 @@ const runUpdateContractsLoop = async ({ signal, provider, abi, contracts }) => {
 
 const runUpdateRewardsLoop = async ({ signal, contracts, ethAddress, onMetrics }) => {
   while (true) {
-    const delay = 10 * 60 * 1000 // 10 minutes
-    const jitter = Math.random() * 20_000 - 10_000 // +- 10 seconds
-    try {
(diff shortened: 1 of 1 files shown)


lib/zinnia.js
```diff
@@ -211,27 +211,26 @@ const runUpdateContractsLoop = async ({ signal, provider, abi, contracts }) => {
...
-      if (err.name === 'AbortError') return
-      throw err
-    }
-
     const contractRewards = await Promise.all(contracts.map(async contract => {
       return getScheduledRewardsWithFallback(contract, ethAddress)
     }))
+    if (signal.aborted) {
+      return
+    }
```
bajtos, line 219: Is it intentional that you are returning before you process `contractRewards`, effectively throwing away the data we expensively fetched?

Maybe it doesn't matter because `lastRewardsScheduledForAddress` isn't used when the signal was aborted, in which case you can ignore my comment.

It also makes me wonder if we should start forwarding `signal` to `getScheduledRewardsWithFallback` so that it can also abort faster.

juliangruber: It doesn't matter whether we use the data here, because when the loop is aborted it starts all over and immediately fetches data again anyway.

I looked into it, unfortunately ethers doesn't support abort signals

PR #401, opened 2024-04-09
change Voyager crashes into `info`

juliangruber:
Screenshot 2024-04-09 at 16 27 57

There's nothing that operators can do about these crashes. This PR changes the activity type into "Info"

### DIFF
--- a/lib/zinnia.js
+++ b/lib/zinnia.js
@@ -260,7 +260,7 @@ const catchChildProcessExit = async ({
       const moduleName = capitalize(err.moduleName ?? 'Zinnia')
       const exitReason = err.exitReason ?? 'for unknown reason'
       const message = `${moduleName} crashed ${exitReason}`
-      const type = moduleName === 'Spark' ? 'error' : 'info'
+      const type = err.moduleName === 'voyager' ? 'info' : 'error'
       onActivity({ type, message })
       const moduleErr = new Error(message, { cause: err })
       // Store the full error message including stdout & stder in the top-level `details` property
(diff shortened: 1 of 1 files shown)


lib/zinnia.js
```diff
@@ -260,7 +260,8 @@ const catchChildProcessExit = async ({
       const moduleName = capitalize(err.moduleName ?? 'Zinnia')
       const exitReason = err.exitReason ?? 'for unknown reason'
       const message = `${moduleName} crashed ${exitReason}`
-      onActivity({ type: 'error', message })
+      const type = moduleName === 'Spark' ? 'error' : 'info'
```
juliangruber, line 263: alternative to consider:

```suggestion
      const type = err.moduleName === 'spark' ? 'error' : 'info'
```

juliangruber: or the other way round, to make the intend more clear:

```suggestion
      const type = err.moduleName === 'voyager' ? 'info' : 'error'
```

bajtos: I like the third option most:
- When we add the third module, its crashes will be reported as errors, which makes more sense to me
- The label "Spark"/"Voyager" can change, but the module codename "spark"/"voyager" should not. Therefore, I prefer to use `err.moduleName` for the check.

bajtos (approved):
TBH, this feels very hacky to me. I prefer to handle these crashes in such a way that no activity logs are created. However, that requires quite a bit of work, starting in Zinnia. In this light, it's best to land this fix as a quick stop-gap solution.

bajtos:
Created a Zinnia feature request here:
- https://github.com/filecoin-station/zinnia/issues/518

juliangruber:
I very much agree!

PR #406, opened 2024-04-12
feat: Generate + store station_id keypair and pass to Zinnia via env var

PatrickNercessian:
Links:
filecoin-station/roadmap#96


lib/cryptoOperations.js
```diff
@@ -0,0 +1,28 @@
+const { subtle } = globalThis.crypto;
+import keytar from 'keytar';
+
+const SERVICE_NAME = 'filecoin-station-core';
+
+export async function getStationId() {
+  const storedKeys = await keytar.getPassword(SERVICE_NAME, 'station_id');
```
juliangruber, line 7: I would like to discuss where to store the private key.

About the [`keytar`](https://github.com/atom/node-keytar) library:
- It is also used in Station Desktop, for storing the wallet seed phrase
- It's deprecated
- On some Windows versions, it's not possible for the user to manually read out the value, without the app offering a function for it

Station Core has two main audiences:
- Inside Station Desktop, running on all 3 major operating systems
- Inside Docker, on some linux flavor

I believe that we want to disallow exporting the private key, as it shouldn't be reused across machines, and doesn't need to be backed up. Therefore, the Windows limitation with keytar is no problem. We can't prevent exporting the private key, since it needs to be stored _somewhere_. But, it's good for us if it isn't easy either.

I believe therefore that keytar _is_ the right solution, as it stores it in a way that's encrypted with the current user's credentials, and isn't as accessible as say a file on disk. I would like to eventually find a replacement for `keytar` that is maintained, but for now we don't have a problem.

One concern I have is that keytar on macOS will ask for the user's password, before allowing operation. Since we use it in Station Desktop and now soon Station Core, we need to prevent asking for the password twice. I think that won't happen, but we need to verify it. I tried running the current PR, but got `[SyntaxError]: Unsupported key usage for an Ed25519 key`. Since it got this far, it seems to indicate that no password confirmation is required. Let's verify this inside Station Desktop once the PR is in a working shape.

commands/station.js
```diff
@@ -5,6 +5,7 @@ import { startPingLoop } from '../lib/telemetry.js'
 import fs from 'node:fs/promises'
 import { metrics } from '../lib/metrics.js'
 import { paths } from '../lib/paths.js'
+import { getStationId } from '../lib/cryptoOperations.js'
```
juliangruber, line 8: what do you think about renaming this to `lib/station-id.js`? I think this more precisely describes the purpose of the file

juliangruber (commented):
Great work so far, @PatrickNercessian! 👏

commands/station.js
PatrickNercessian: Works for me!

lib/cryptoOperations.js
PatrickNercessian: This all makes sense to me!

> I believe that we want to disallow exporting the private key

Do you mean we should do something extra to disallow exporting the private key here, or just that keytar is a good solution for now because it makes this nontrivial?

> Let's verify this inside Station Desktop once the PR is in a working shape.

Do you recommend that I try to get my own dev application working on my machine to do this spot-checking, or that you and Miro do this, and I focus on unit tests and such?

juliangruber: > Do you mean we should do something extra to disallow exporting the private key here, or just that keytar is a good solution for now because it makes this nontrivial?

The latter 👍 

> Do you recommend that I try to get my own dev application working on my machine to do this spot-checking, or that you and Miro do this, and I focus on unit tests and such?

I'm happy to do it, just ping me 🙏

commands/station.js
```diff
@@ -82,6 +83,7 @@ export const station = async ({ json, experimental }) => {
 
   const modules = [
     zinniaRuntime.run({
+      STATION_ID: (await getStationId()).publicKey,
```
bajtos, line 86: If calling keytar is going to trigger a UI prompt on macOS, I strongly prefer to make this call only once during station-core startup.

I am also proposing to include station id in the `ping` telemetry we are submitting here, which is yet another argument for moving the `getStationId()` call somewhere else.

https://github.com/filecoin-station/core/blob/f8fd023d90715202ef5d23d253776b8036ee6456/lib/telemetry.js#L52-L56

bajtos: Hmm, I think I got confused. This file is called only once, and then it's `zinniaRuntime.run` that handles module updates and (re) starts the runtime.

In that light, it's okay to call `getStationId()` inside the `station()` function.

However, I would find the code cleaner if this call were happening somewhere closer to the top, where we are also reading FIL_WALLET_ADDRESS, and the ID was stored in a variable.

```js
const { publicKey: STATION_ID } = await getStationId()
// ...
  const modules = [
    zinniaRuntime.run({
      STATION_ID,
      // ...
```

bajtos:
@PatrickNercessian would you mind adding Station ID to the `ping` telemetry points?

https://github.com/filecoin-station/core/blob/f8fd023d90715202ef5d23d253776b8036ee6456/lib/telemetry.js#L52-L56

Proposed change in `commands/station.js`

```diff
-  startPingLoop().unref()
+  startPingLoop({ STATION_ID }).unref()
```

commands/station.js
```diff
@@ -22,6 +23,8 @@ const panic = msg => {
 }
 
 export const station = async ({ json, experimental }) => {
+  const STATION_ID = (await getStationId()).publicKey
```
juliangruber, line 26: @bajtos do you have thoughts on whether UPPER_CASE should be reserved for environment variables, or is it for constants in general?

bajtos: I don't have a strong opinion. 

We are using UPPER_CASE for many fields in the options parameters passed around, so I thought it would be more consistent to use `STATION_ID` here. (Compare this name to `FIL_WALLET_ADDRESS` referenced in this file.)

IMO, variable names should describe what value is stored in them. How we initialise the name is a detail that can vary, therefore it should not influence the name. For example, in tests, we often set FIL_WALLET_ADDRESS to a hard-coded value.

Personally, I would do this:

```js
const {
  FIL_WALLET_ADDRESS: walletAddress,
  // ...
} = process.env

// then use `walletAddress` everywhere else
```

juliangruber: Ok! Great that you value consistency more!

> Personally, I would do this:

Thinking about it some, I would also prefer this. Let's address it in the future if we feel strong enough about it

bajtos (approved):
LGTM 👍🏻 

@juliangruber: Could you please check the new keytar interaction when both Desktop and Core read from the keychain at startup?

juliangruber:
Tested and works 👍

bajtos:
@SocketSecurity ignore npm/keytar@7.9.0

bajtos:
The builds are failing on Linux now.

```
[Error: The name org.freedesktop.secrets was not provided by any .service files]
```

```
node:internal/modules/cjs/loader:1465
  return process.dlopen(module, path.toNamespacedPath(filename));
                 ^

Error: libsecret-1.so.0: cannot open shared object file: No such file or directory
    at Module._extensions..node (node:internal/modules/cjs/loader:1465:18)
    at Module.load (node:internal/modules/cjs/loader:1206:32)
    at Module._load (node:internal/modules/cjs/loader:1022:12)
    at Module.require (node:internal/modules/cjs/loader:1231:19)
    at require (node:internal/modules/helpers:179:18)
    at Object.<anonymous> (/usr/src/app/node_modules/keytar/lib/keytar.js:1:14)
    at Module._compile (node:internal/modules/cjs/loader:1369:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1427:10)
    at Module.load (node:internal/modules/cjs/loader:1206:32)
    at Module._load (node:internal/modules/cjs/loader:1022:12) {
  code: 'ERR_DLOPEN_FAILED'
}
```

bajtos:
I added [0fe59a9](https://github.com/filecoin-station/core/pull/406/commits/0fe59a9e50032ae717cb58d4bffc5b35011637a9) to install `libsecret` required by `node-tar`.

We are getting a different error now:

```
[Error: Object does not exist at path “/org/freedesktop/secrets/collection/login”]
```

@PatrickNercessian could you please look into this yourself?

Maybe we can copy the setup steps from keytar's CI, see the following two blocks:
- https://github.com/atom/node-keytar/blob/deae59a488789f2cd4a8dba6c7e58665795804fe/.github/workflows/ci.yml#L39-L45
- https://github.com/atom/node-keytar/blob/deae59a488789f2cd4a8dba6c7e58665795804fe/.github/workflows/ci.yml#L62-L73

That code is more than two years old; there may be easier options available by now. Feel free to use a different approach to fix this problem!

juliangruber:
See https://github.com/PatrickNercessian/core/pull/1

PR #409, opened 2024-04-16
modules: add timeout fetching module sources

juliangruber:


juliangruber:
@bajtos hindsight please, need to release as without this the gateway fallback breaks for some

PR #410, opened 2024-04-16
Fix pipeline types

juliangruber:
Tests are failing locally, but not in CI. This is because we didn't include the `build` step in CI. This PR fixes that. It also then disables it temporarily, so that we can publish new versions of this project again.

@bajtos I wasn't able to figure out the type failures unfortunately :| Do you have an idea?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,6 @@ jobs:
           node-version: ${{ matrix.node }}
       - run: npm ci --omit=dev
       - run: npm ci
-      - run: npm run test:types
       - run: npm run test:lint
         if: matrix.os != 'windows-latest'
       - run: npm run test:unit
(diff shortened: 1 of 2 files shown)


juliangruber:
Hindsight review please, I need to get an urgent release out

bajtos:
> @bajtos I wasn't able to figure out the type failures unfortunately :| Do you have an idea?

When I oversimplify the problem:
- `pipeline` expects the first argument to be Node.js `ReadableStream` and the second argument to be Node.js `WritableStream`
- `res.body` has an Unidici-specific readable-stream type that's similar to Node.js readable-stream type but not fully compatible from TypeScript's point of view
- `tar.x` returns tar-specific type `Unpack`, which is similar to Node.js writable-stream but not fully compatible from TypeScript's point of view

This could be also a problem of `@types/node` or the typings provided by Undici or Tar.

I'll open a PR to fix this.

PR #411, opened 2024-04-16
Add machine info to telemetry. Closes #404

### LINKED ISSUE #404: Capture machine metrics
In order to better know how much resources the network is able to offer, capture:

- [ ] CPU
  - [ ] core count?
  - [ ] core clock?
  - [ ] core brand?
  - [ ] core model?
- [ ] Storage: free/total
- [ ] Memory: free/total
- [x] OS https://github.com/filecoin-station/desktop/pull/582 https://github.com/filecoin-station/core/pull/49

juliangruber:
Closes #404

I didn't see a good way to test free disk space (depends on mount points etc), but that's not relevant for Station modules atm anyway.

### DIFF
--- a/lib/telemetry.js
+++ b/lib/telemetry.js
@@ -87,7 +87,6 @@ export const runMachinesLoop = async () => {
             ? 'apple'
             : 'unknown'
     )
-    point.intField('memory_free_b', freemem())
     point.intField('memory_total_b', totalmem())
     writeClientMachines.writePoint(point)
     writeClientMachines.flush().catch(handleFlushError)
--- a/lib/telemetry.js
+++ b/lib/telemetry.js
@@ -76,13 +76,16 @@ export const runMachinesLoop = async () => {
     const point = new Point('machine')
     point.intField('cpu_count', cpus().length)
     point.intField('cpu_speed_mhz', cpus()[0].speed)
+    const model = cpus()[0].model.toLowerCase()
     point.tag(
       'cpu_brand',
-      /intel/i.test(cpus()[0].model)
+      model.includes('intel')
         ? 'intel'
-        : /amd/i.test(cpus()[0].model)
+        : model.includes('amd')
           ? 'amd'
-          : 'unknown'
+          : model.includes('apple')
+            ? 'apple'
+            : 'unknown'
     )
     point.intField('memory_free_b', freemem())
     point.intField('memory_total_b', totalmem())
--- a/lib/telemetry.js
+++ b/lib/telemetry.js
@@ -93,6 +93,8 @@ export const runMachinesLoop = async () => {
         point.stringField('cpu_model_unknown_brand', model)
       }
     }
+    point.tag('platform', platform())
+    point.tag('arch', arch())
     point.intField('memory_total_b', totalmem())
     writeClientMachines.writePoint(point)
     writeClientMachines.flush().catch(handleFlushError)
--- a/lib/telemetry.js
+++ b/lib/telemetry.js
@@ -39,22 +39,20 @@ const writeClientMachines = client.getWriteApi(
 const unactionableErrors =
   /HttpError|getAddrInfo|RequestTimedOutError|ECONNRESET|EPIPE|ENETDOWN|ENOBUFS|EHOSTUNREACH|ERR_TLS_CERT_ALTNAME_INVALID|ETIMEDOUT|EPROTO|ENETUNREACH/i
 
-setInterval(() => {
-  writeClient.flush().catch(err => {
-    if (unactionableErrors.test(String(err))) {
-      return
-    }
-    if (typeof err?.code === 'string' && unactionableErrors.test(err.code)) {
-      return
-    }
-    Sentry.captureException(err)
-  })
-}, 5_000).unref()
+const handleFlushError = err => {
+  if (unactionableErrors.test(String(err))) {
+    return
+  }
+  if (typeof err?.code === 'string' && unactionableErrors.test(err.code)) {
+    return
+  }
+  Sentry.captureException(err)
+}
 
-export const startPingLoop = () => {
+export const runPingLoop = async () => {
   assert(FIL_WALLET_ADDRESS)
   const processUUID = randomUUID()
-  return setInterval(() => {
+  while (true) {
     const point = new Point('ping')
     point.stringField(
       'wallet',
@@ -67,31 +65,28 @@ export const startPingLoop = () => {
     point.tag('arch', arch())
     point.tag('deployment_type', DEPLOYMENT_TYPE)
     writeClient.writePoint(point)
-  }, 10 * 60 * 1000)
+    writeClient.flush().catch(handleFlushError)
+    await timers.setTimeout(10 * 60 * 1000) // 10 minutes
+  }
 }
 
 export const runMachinesLoop = async () => {
   while (true) {
-    try {
-      const point = new Point('machine')
-      point.intField('cpu_count', cpus().length)
-      point.intField('cpu_speed_mhz', cpus()[0].speed)
-      point.tag(
-        'cpu_brand',
-        /intel/i.test(cpus()[0].model)
-          ? 'intel'
-          : /amd/i.test(cpus()[0].model)
-            ? 'amd'
-            : 'unknown'
-      )
-      point.intField('memory_free_b', freemem())
-      point.intField('memory_total_b', totalmem())
-      writeClientMachines.writePoint(point)
-      await writeClientMachines.flush()
-    } catch (err) {
-      console.error(err)
-      Sentry.captureException(err)
-    }
+    const point = new Point('machine')
+    point.intField('cpu_count', cpus().length)
+    point.intField('cpu_speed_mhz', cpus()[0].speed)
+    point.tag(
+      'cpu_brand',
+      /intel/i.test(cpus()[0].model)
+        ? 'intel'
+        : /amd/i.test(cpus()[0].model)
+          ? 'amd'
+          : 'unknown'
+    )
+    point.intField('memory_free_b', freemem())
+    point.intField('memory_total_b', totalmem())
+    writeClientMachines.writePoint(point)
+    writeClientMachines.flush().catch(handleFlushError)
     await timers.setTimeout(86_400) // 1 day
   }
 }
--- a/lib/telemetry.js
+++ b/lib/telemetry.js
@@ -2,9 +2,10 @@ import { InfluxDB, Point } from '@influxdata/influxdb-client'
 import { createHash, randomUUID } from 'node:crypto'
 import Sentry from '@sentry/node'
 import assert from 'node:assert'
-import { platform, arch } from 'node:os'
+import { platform, arch, cpus, freemem, totalmem } from 'node:os'
 import fs from 'node:fs/promises'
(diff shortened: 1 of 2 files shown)


lib/telemetry.js
```diff
@@ -60,5 +66,28 @@ export const startPingLoop = () => {
...
+    point.intField('cpu_count', cpus().length)
+    point.intField('cpu_speed_mhz', cpus()[0].speed)
+    point.tag(
+      'cpu_brand',
+      /intel/i.test(cpus()[0].model)
+        ? 'intel'
+        : /amd/i.test(cpus()[0].model)
+          ? 'amd'
+          : 'unknown'
+    )
```
bajtos, line 86: Please add code to detect ARM, especially Apple Silicon (M1/M2/M3).

lib/telemetry.js
```diff
@@ -60,5 +66,28 @@ export const startPingLoop = () => {
...
+    point.intField('memory_free_b', freemem())
```
bajtos, line 87: The amount of free memory changes often, but we are sampling the value only once per day. I think such data would most likely mislead us.

I slightly prefer to remove this field, but I don't feel too strongly about that.

lib/telemetry.js
```diff
@@ -60,5 +66,28 @@ export const startPingLoop = () => {
...
     writeClient.writePoint(point)
-  }, 10 * 60 * 1000)
+    writeClient.flush().catch(handleFlushError)
+    await timers.setTimeout(10 * 60 * 1000) // 10 minutes
+  }
+}
+
+export const runMachinesLoop = async () => {
+  while (true) {
+    const point = new Point('machine')
```
bajtos, line 76: How can we distinguish points submitted by different machines? I don't see any machine-specific identifier in the fields.

bajtos (commented):
@juliangruber could you please coordinate with @PatrickNercessian? I think he will want to use this data for the public dashboard too.

lib/telemetry.js
juliangruber: Hm, why do we need to distinguish points by different machines?

lib/telemetry.js
juliangruber: I agree, let's remove it for now

lib/telemetry.js
juliangruber: I've added ARM support for apple silicon. What other CPU brands should we detect?

juliangruber:
> @juliangruber could you please coordinate with @PatrickNercessian? I think he will want to use this data for the public dashboard too.

Good idea! @PatrickNercessian what do you think about these machine metrics and how they are stored?

lib/telemetry.js
bajtos: > What other CPU brands should we detect?

IDK 🤷🏻‍♂️ Maybe you can query our current Station `ping` events to see what other platforms we have seen so far?

lib/telemetry.js
bajtos: That's a good question. How do you envision to query this new data?

I assume you want to write a query like "find me a sum of all RAM sizes of active machines last week". (In other words, find how much RAM is collectively offered by our network.)

Let's say Station A with 4GB of RAM reports a telemetry point every day (Mon, Tue,..., Sun). Station B, with 1GM of RAM, reports only one telemetry point on Wednesday. Station C, with 2GB of RAM, reports two telemetry points on Sat & Sun.

How are you going to calculate the correct answer 8 GB?

We may also want to know the max & min peaks (4 GB on a bad day, 8 GB on a good day) and perhaps the p50 value (a bit more than 4 GB).

lib/telemetry.js
```diff
@@ -60,5 +66,30 @@ export const startPingLoop = () => {
...
+        : model.includes('amd')
+          ? 'amd'
+          : model.includes('apple')
+            ? 'apple'
+            : 'unknown'
+    )
+    point.intField('memory_total_b', totalmem())
+    writeClientMachines.writePoint(point)
+    writeClientMachines.flush().catch(handleFlushError)
+    await timers.setTimeout(86_400) // 1 day
```
bajtos, line 93: I think this is 1/1000 of a day if `setTimeout` accepts the interval in milliseconds.

Do you know if the implementation in Node.js tracks wall time or process time? When my computer sleeps for 8 hours during the night, will the timer keep progressing, or will it be delayed by that 8 hours?

Maybe we can submit the machine stats more often, e.g. every 4 hours or every 6 hours?

PatrickNercessian:
> > @juliangruber could you please coordinate with @PatrickNercessian? I think he will want to use this data for the public dashboard too.
> 
> Good idea! @PatrickNercessian what do you think about these machine metrics and how they are stored?

The original plan was to use Rust's built-in libraries for device info like this in Zinnia, and pass it to the modules. Do we prefer to check this inside Core and pass it to Zinnia who then passes it to the modules, similar to station_id?

If so, should we instead generate this info outside of the telemetry context, so we can pass it Zinnia easily, and then just pass it to telemetry functions?

bajtos:
I am strongly against passing machine info from Station Core to Zinnia and Station Modules.

As I understand the proposal in this pull request, it records machine telemetry directly from Station Core to our InfluxDB, as an alternative design/solution to what we envisioned with Zinnia & Station Modules.

If we land this PR, we will have telemetry points about Station machines written to InfluxDB and stored in a bucket with a 30-day retention (see https://github.com/filecoin-station/core/issues/404).

How much work is it to ingest this telemetry from InfluxDB to the Station public dashboard you are building, @PatrickNercessian? I think this will require a new component that is different from what we have discussed so far. I think it will need to periodically read data in InfluxDB and write aggregated metrics to a place where they can be persisted forever and exposed for the Dashboard UI.

lib/telemetry.js
juliangruber: Platform != CPU brand though. Do you think we should submit platform in addition to CPU brand?

lib/telemetry.js
juliangruber: I would say it depends on what query we want to ask. But I see that I can't harm to add a machine ID, to support more queries. I'll add it 👍

juliangruber: 0cd2668

juliangruber:
I agree with @bajtos, I don't see the benefit in passing this module data to the module source. Every Station should submit this, and modules don't need to be aware. Also, this is not per-module information. Or am I missing why this is important for the module to read?

juliangruber:
I think your plan might have been to submit this machine info with every measurement that the module submits, right? Then indeed it would need to be passed to the module.

lib/telemetry.js
bajtos: I don't know what kind of information we want to get out of this telemetry, so it's hard to answer, TBH.

> What other CPU brands should we detect?

What CPU brands are used by home servers like Raspberry PI?

> Do you think we should submit platform in addition to CPU brand?

I think there is no harm in doing that.

Alternatively, instead of submitting `unknown` brand, we can include the platform in that value, e.g. `unknown-x86`, `unknown-x64`, `unknown-arm64`, etc.

lib/telemetry.js
```diff
@@ -60,5 +67,31 @@ export const startPingLoop = () => {
...
+        : model.includes('amd')
+          ? 'amd'
+          : model.includes('apple')
+            ? 'apple'
+            : 'unknown'
+    )
+    point.intField('memory_total_b', totalmem())
+    writeClientMachines.writePoint(point)
+    writeClientMachines.flush().catch(handleFlushError)
+    await timers.setTimeout(86_400 * 1000) // 1 day
```
bajtos, line 95: Nitpick, feel free to ignore.

```suggestion
    await timers.setTimeout(86_400_000) // 1 day
```

The ping loop uses multiplication instead:

```js
await timers.setTimeout(10 * 60 * 1000) // 10 minutes
```

Maybe we should do the same in this place for consistency.

```suggestion
    await timers.setTimeout(24 * 3600 * 1000) // 1 day
```

bajtos:
BTW, in the longer term, I'd like to replace the process-uuid value with `stationId` introduced by #406, sign the payload using the private key and include the signature as another field in the telemetry point. This allows consumers to verify that the correct Station Core instance submitted the machine spec, making it more difficult for potential attackers to submit a fake machine spec on behalf of a victim.

By including `stationId` in the machine spec, we can also link measurements committed to Spark or Voyager to the specs of the machine that performed those measurements.

I am just sharing my thoughts and visions. This is not blocking progress on this pull request; we can land it without `stationId`.

lib/telemetry.js
juliangruber: I don't have access to a Respberry PI, so could only guess :|

> Alternatively, instead of submitting unknown brand, we can include the platform in that value, e.g. unknown-x86, unknown-x64, unknown-arm64, etc.

Good idea! I've improved this even further. Wdyt?

PatrickNercessian:
Ah I didn't realize this was writing to our centralized InfluxDB, I assumed it was for something locally in each Core deployment...

I haven't worked with InfluxDB before, so the reliability of my estimate of effort is somewhat limited. That being said, I can think of a few routes we could go.

A) In spark-evaluate, for every stationId within every round (we don't need to do for every measurement), we check InfluxDB for the latest machine info, and store that in the (stationId, day) row of `daily_node_metrics` along with the rest of the columns. Whether or not this makes sense might depend on how long current round evaluations currently take, and how long the latency would be for checking the InfluxDB for this information
B) In some new continuously polling service, we check InfluxDB for new machine information, and store it in a new Postgres table. Then the `daily_node_metrics` table points to that table's information as a foreign key

There are probably some other ways we could do it too, these are my first thoughts

lib/telemetry.js
```diff
@@ -60,5 +67,37 @@ export const startPingLoop = () => {
...
-  }, 10 * 60 * 1000)
+    writeClient.flush().catch(handleFlushError)
+    await timers.setTimeout(10 * 60 * 1000) // 10 minutes
+  }
+}
+
+export const runMachinesLoop = async () => {
+  while (true) {
+    const point = new Point('machine')
+    point.stringField('process_uuid', processUUID)
```
bajtos, line 78: Can you please add `station_id` to these data points too? (Alternatively, remove `process_uuid` and use `station_id` only.)

juliangruber:
@PatrickNercessian we discussed requirements for this and realized that storing in Influx is fine, since we are mostly interested at the _current_ state of the network, not historic data, and that is easy and cheap to query from Influx

PatrickNercessian:
> @PatrickNercessian we discussed requirements for this and realized that storing in Influx is fine, since we are mostly interested at the _current_ state of the network, not historic data, and that is easy and cheap to query from Influx

Sounds good. Do our centralized services have permission to read from InfluxDB? I don't see any existing code about it.

Also, any thoughts on the above two ideas about high-level implementation for this querying?

bajtos (approved):
LGTM

juliangruber:
> Sounds good. Do our centralized services have permission to read from InfluxDB? I don't see any existing code about it.

Yes! With an access token, it's readable from the public internet. I believe there were some snippets reading from it, currently our Grafana instance is the main consumer.

> Also, any thoughts on the above two ideas about high-level implementation for this querying?

I think it would be nicer for spark-stats to periodically query the information. It can also listen to the smart contract for round start events. Otherwise we add a task to spark-evaluate that has nothing to do with evaluating the work done.

PR #414, opened 2024-04-18
deps: upgrade Zinnia to v0.18.1

bajtos:
https://github.com/filecoin-station/zinnia/releases/tag/v0.18.1

Blocked by:
- #406

Links:
- filecoin-station/roadmap#96

PR #424, opened 2024-04-23
feat: persisted Station ID passed to Zinnia

bajtos:
Introduce the concept of a Station instance ID that's represented as a public key of an Ed25119 keypair.

Generate the keypair on the first start and store it in a state file. Load the keypair from the state file on subsequent starts.

Introduce an optional env var PASSPHRASE allowing operators to specific a passphrase that will be used to encrypt & decrypt the file storing the keypair.

Implement an upgrade path allowing users to start with an empty PASSPHRASE and later upgrade their deployment with a PASSPHRASE string. Station will load the plaintext keypair and encrypt it using the new PASSPHRASE when it detects this situation.

Forward the Station ID to Zinnia via env vars and include it in the telemetry data written to InfluxDB.

Closes #406


README.md
```diff
@@ -81,6 +81,9 @@ the configuration options described in
   `0x000000000000000000000000000000000000dEaD`. Please note that any earnings
   sent there will be lost.
 
+- `PASSPHRASE` _(string; required)_: a passphrase to protect the Station
+  instance private key stored in a file inside `STATE_ROOT` directory.
```
bajtos, line 85: Adding a new required env var is a breaking change. The next release will break all Core deployments using automatic upgrades (e.g. via Watchtower for Docker).

It's the least amount of work for us. However, I am concerned about the impact on our users.

How can we make this change backwards compatible?

One option that comes to my mind:
- If `PASSPHRASE` is not provided, then use one of the following fallbacks (we should pick & implement only one of them):
  1. Use the FIL_WALLET_ADDRESS as the passphrase
  2. Use a hard-coded passphrase
  3. Don't encrypt the private key at all
- When persisting the keypair, add a flag indicating whether we used the fallback or the user-provided passphrase.
- When loading the keypair, check the flag to decide how to decrypt the persisted keypair.
- When loading the keypair encrypted using the fallback mechanism when the user provided the passphrase, then upgrade our secret store to encrypt the keypair using the provided passphrase.
- When loading the keypair using the fallback mechanism, print a deprecation warning asking the user to configure their own PASSPHRASE.

It's a non-trivial amount of work to achieve backwards compatibility. Is it worth it? Can we find a simpler way?

PatrickNercessian: Instead of flagging the use of a fallback versus user-provided passphrase, can we treat them as one and the same, and provide a way for the user to update their passphrase at any time?

Then we can remove these two bullet points of work:
~When persisting the keypair, add a flag indicating whether we used the fallback or the user-provided passphrase.~
~When loading the keypair, check the flag to decide how to decrypt the persisted keypair.~

And we print the deprecation warning (if we want to at all) based on if the passphrase (hash) matches the fallback (hash) that we chose

bajtos: > provide a way for the user to update their passphrase at any time

Interesting! How do you propose to implement this? A new sub-command for station core? E.g. `station edit-passphrase`? That could work 🤔

PatrickNercessian: That's what I was imagining, although I haven't used core so I didn't know the specifics enough to mention it as a suggestion.

What do we want to do for Desktop then? Do we need UI changes to allow this functionality? Or should we just use keytar for Desktop deployments, since it already uses it?

bajtos: > What do we want to do for Desktop then? Do we need UI changes to allow this functionality? Or should we just use keytar for Desktop deployments, since it already uses it?

I discussed this with Julian, and we are thinking of deriving a passphrase from the wallet seed. We could add another keychain item via keytar, but that would add another prompt on macOS, which I don't like.

The idea is to derive the passphrase in such a way that the passphrase is unique enough & difficult to guess but also does not reveal too much of the seed phrase to the Station Core, in case there is a security vulnerability there.

.github/workflows/ci.yml
```diff
@@ -90,6 +90,7 @@ jobs:
            --name station \
            --detach \
            --env FIL_WALLET_ADDRESS=0x000000000000000000000000000000000000dEaD \
+           --env PASSPHRASE=secret \
```
juliangruber, line 93: A passphrase is a collection of words, while a password is any string of characters. Was it intentional do select a passphrase here?

README.md
juliangruber: > Adding a new required env var is a breaking change. The next release will break all Core deployments using automatic upgrades (e.g. via Watchtower for Docker).
>
> It's the least amount of work for us. However, I am concerned about the impact on our users.

What about making it optional instead? It's also optional with ssh. Optional but advised

When a passphrase isn't used, the private key is not encrypted at all. When you start passing one, it will fail and tell the user to delete the private key located at PATH, if they want to set a new passphrase.

This way we achieve backwards compatibility, give users an easy upgrade path, and just have 1 key regeneration per node max.

lib/paths.js
```diff
@@ -13,6 +13,7 @@ const {
 } = process.env
 
 export const getPaths = ({ cacheRoot, stateRoot }) => ({
+  secretStore: join(stateRoot, 'secrets'),
```
juliangruber, line 16: Didn't you say in our call this should be in cacheRoot, because stateRoot might be shared between machines? Or is this path safe after all?

juliangruber:
```suggestion
  secrets: join(stateRoot, 'secrets'),
```

To me, `paths.secrets` is sufficient. `secretsStore` sounds like an abstraction for storing secrets.

lib/station-id.js
```diff
@@ -0,0 +1,142 @@
...
+import { subtle, getRandomValues } from 'node:crypto'
+
+/**
+ * @param {object} args
+ * @param {string} args.secretStore
+ * @param {string} args.passphrase
+ * @returns {Promise<{publicKey: string, privateKey: string}>}
+ */
+export async function getStationId ({ secretStore, passphrase }) {
+  assert.strictEqual(typeof secretStore, 'string', 'stateRoot must be a string')
```
juliangruber, line 13: variable name missmatch here

juliangruber (changes requested):

README.md
bajtos: > > Adding a new required env var is a breaking change. The next release will break all Core deployments using automatic upgrades (e.g. via Watchtower for Docker).
> > It's the least amount of work for us. However, I am concerned about the impact on our users.
> 
> What about making it optional instead? It's also optional with ssh. Optional but advised
> 
> When a passphrase isn't used, the private key is not encrypted at all. When you start passing one, it will fail and tell the user to delete the private key located at PATH, if they want to set a new passphrase.
> 
> This way we achieve backwards compatibility, give users an easy upgrade path, and just have 1 key regeneration per node max.

That's an option, too.

When the user regenerates the key, the Station ID will change. We will count one station instance as two instances in that month. That's why I prefer to implement an automatic update of the keypair file when the passphrase changes.

If we are fine with the possibility that Station ID can infrequently change and thus we will incorrectly count the number of unique instances, then I am proposing a small change to your proposal:
- When the user starts providing a passphrase, we automatically delete the old file and generate a new one.
- We can broaden the conditions triggering new keypair creation to all cases where the file cannot be decrypted using the provided passphrase. This way, the Station Core operator does not need to perform any manual steps related to the passphrase. The downside is that whenever the passphrase changes, so does the Station ID, and therefore, we can double-count station instances more often than we would like to.

.github/workflows/ci.yml
bajtos: SSH and GPG use the term "passphrase", so I chose it here.
- https://docs.github.com/en/authentication/connecting-to-github-with-ssh/working-with-ssh-key-passphrases
- https://www.gnupg.org/gph/en/manual/c14.html

TBH, I find the env var name PASSPHRASE a bit ambiguous, so I am happy to use a different name if we can find a better one. What would you suggest?

lib/paths.js
bajtos: > Didn't you say in our call this should be in cacheRoot, because stateRoot might be shared between machines? Or is this path safe after all?

I did say that. Then I looked into `paths.js` and found that we are using `APPLOCALDATA` for storing the state, which means the state stays local - that's exactly what we need.

I think it would be great to capture the requirement for `stateRoot` to be local and not shared with other computers the operator may use, but I don't know how. We can add code comments (where?) and mention this requirement in README too. WDYT?

README.md
juliangruber: > When the user regenerates the key, the Station ID will change. We will count one station instance as two instances in that month. That's why I prefer to implement an automatic update of the keypair file when the passphrase changes.

Yes, but we only have 1 double counting per node max, one time. I don't think that's particularly bad and worth more complexity

juliangruber: > When the user starts providing a passphrase, we automatically delete the old file and generate a new one.

How do we detect when a user starts providing a passphrase? We can only tell that a passphrase was provided, but not that one wasn't provided before, right? Or can we deduct this information from the file?

If so, what about instead of regenerating the private key, we simply take the unencrypted value, encrypt it, and overwrite the file? This way the operator doesn't notice and the station id doesn't change.

juliangruber: If this automatic process isn't possible, I'm fully in favor of making the core operator manage this. They're most likely a linux admin level user.

.github/workflows/ci.yml
juliangruber: Cool, if they use this, then I'm all in favor.

> TBH, I find the env var name PASSPHRASE a bit ambiguous, so I am happy to use a different name if we can find a better one. What would you suggest?

Explicit would be `KEY_PASSPHRASE` or `PRIVATE_KEY_PASSPHRASE` / `PKEY_PASSPHRASE`

lib/paths.js
juliangruber: Good idea to add a comment in paths.js and also mention this briefly in the README (I don't think many users will run into this situation where it's shared)

README.md
bajtos: I think it's easy to detect if the file is not encrypted - it will contain plaintext JSON starting with the character `{`. The encrypted file contains binary data.

I'll update this PR with the following changes:
- Make the PASSPHRASE optional
- When PASSPHRASE is not provided, we store the keypair in plaintext
- When loading the keypair, handle the upgrade path from unencrypted to encrypted

.github/workflows/ci.yml
bajtos: We are encrypting the entire keypair now - `JSON.stringify({privateKey, publicKey})`.

Should we call the env var `KEYPAIR_PASSPHRASE` then? It's a bit long, but comparable to `FIL_WALLET_ADDRESS`.

OTOH, I can imagine we can reuse this passphrase for other things in the future, so maybe it's better to keep the generic `PASSPHRASE` name.

README.md
bajtos: > I think it's easy to detect if the file is not encrypted - it will contain plaintext JSON starting with the character `{`. The encrypted file contains binary data.
> 
> I'll update this PR with the following changes:
> 
> * Make the PASSPHRASE optional
> * When PASSPHRASE is not provided, we store the keypair in plaintext
> * When loading the keypair, handle the upgrade path from unencrypted to encrypted

Implemented in [a9babf9](https://github.com/filecoin-station/core/pull/424/commits/a9babf9e8a44442e9f13bec51cf780d4c80347f2)

.github/workflows/ci.yml
bajtos: Let's keep using `PASSPHRASE`.

lib/paths.js
bajtos: Addressed in [c139c6c](https://github.com/filecoin-station/core/pull/424/commits/c139c6c90912d197a7c45909635d02528a3e8411)

bajtos:
@juliangruber @PatrickNercessian I addressed the comments, and the pull request is ready for final review and merging.

@juliangruber if we manage to land this PR today, could you please land https://github.com/filecoin-station/core/pull/414 as well, before publishing a new Station Core release.

lib/station-id.js
```diff
@@ -0,0 +1,206 @@
...
+import { subtle, getRandomValues } from 'node:crypto'
+
+/**
+ * @param {object} args
+ * @param {string} args.secretsDir
+ * @param {string} args.passphrase
+ * @returns {Promise<{publicKey: string, privateKey: string}>}
+ */
+export async function getStationId ({ secretsDir, passphrase }) {
+  assert.strictEqual(typeof secretsDir, 'string', 'stateRoot must be a string')
```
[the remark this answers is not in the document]
PatrickNercessian: Approving PR anyway, since this is a nitpick, and we want to land ASAP

PatrickNercessian (approved):

lib/station-id.js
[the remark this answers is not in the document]
bajtos: oh, I thought I have fixed this 🙈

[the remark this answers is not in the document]
bajtos: Fixed in [3e6fe1e](https://github.com/filecoin-station/core/pull/424/commits/3e6fe1eca54e38bf0ce6e29e5d460291da11cba0)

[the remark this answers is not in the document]
bajtos: Thank you, @PatrickNercessian, for catching this!

README.md
```diff
@@ -54,10 +54,16 @@ The following configuration options are shared by all Station commands:
   - Windows: `%TEMP%/Filecoin Station Core`
 - `$STATE_ROOT` _(string; optional)_: Station stores logs and module state in
   this directory. Defaults to
+
   - Linux: `${XDG_STATE_HOME:-~/.local/state}/filecoin-station-core`
   - macOS: `~/Library/Application Support/app.filstation.core`
   - Windows: `%LOCALAPPDATA%/Filecoin Station Core`
 
+  **IMPORTANT:** State files must be stored in a directory that's local to the
```
juliangruber, line 62:
```suggestion
  **IMPORTANT:** The`$STATE_ROOT` directory must be local to the
```

juliangruber: As a reader I wouldn't be certain whether by "State files" we mean that directory

juliangruber (approved):

PR #433, opened 2024-04-26
zinnia: add restart module on inactivity

juliangruber:


lib/zinnia.js
```diff
@@ -365,8 +365,22 @@ export async function run ({
...
+      if (timeoutId) clearTimeout(timeoutId)
+      timeoutId = setTimeout(() => {
+        onActivity({
+          type: 'error',
+          message: `${capitalize(module)} has been inactive for 5 minutes, restarting...`
+        })
+        controller.abort()
+      }, 5 * 60 * 1000)
+    }
+    let timeoutId
```
bajtos, line 378: Would you mind moving this line before `resetTimeout` so that `resetTimeout` does not use a variable that was defined yet (from the POV of a reader)?

lib/zinnia.js
```diff
@@ -365,8 +365,22 @@ export async function run ({
     )
     childProcesses.push(Object.assign(childProcess, { moduleName: module }))
 
+    const resetTimeout = () => {
+      if (timeoutId) clearTimeout(timeoutId)
```
bajtos, line 369: I think we need to clear the timeout after `Zinnia main loop ended`, otherwise we will end up with a leaked/danging timeout after restarting the main loop e.g. after a new module version was detected.

bajtos (changes requested):
Great idea! Let's improve the implementation details.

bajtos (approved):
👌🏻

PR #434, opened 2024-04-26
zinnia: fix double reporting of crash errors

juliangruber:
- [x] swap sentry token

### DIFF
--- a/lib/zinnia.js
+++ b/lib/zinnia.js
@@ -53,7 +53,7 @@ const maybeReportErrorToSentry = (/** @type {unknown} */ err) => {
   /** @type {Parameters<Sentry.captureException>[1]} */
   const hint = { extra: {} }
   if (typeof err === 'object') {
-    if ('reportedToSentry' in err && err.reportToSentry === true) {
+    if ('reportedToSentry' in err && err.reportedToSentry === true) {
       return
     }
     Object.assign(err, { reportedToSentry: true })
--- a/lib/zinnia.js
+++ b/lib/zinnia.js
@@ -53,10 +53,10 @@ const maybeReportErrorToSentry = (/** @type {unknown} */ err) => {
   /** @type {Parameters<Sentry.captureException>[1]} */
   const hint = { extra: {} }
   if (typeof err === 'object') {
-    if ('reportToSentry' in err && err.reportToSentry === false) {
+    if ('reportedToSentry' in err && err.reportToSentry === true) {
       return
     }
-    Object.assign(err, { reportToSentry: false })
+    Object.assign(err, { reportedToSentry: true })
     if ('details' in err && typeof err.details === 'string') {
       // Quoting from https://develop.sentry.dev/sdk/data-handling/
       // > Messages are limited to 8192 characters.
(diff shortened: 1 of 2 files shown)


lib/zinnia.js
```diff
@@ -276,6 +279,8 @@ const catchChildProcessExit = async ({
           maybeReportErrorToSentry(moduleErr)
         })
       }
+      // If it should be reported, it was already handled
+      err.reportToSentry = false
```
bajtos, line 283: I am a bit confused about why we are reporting the error to Sentry twice. Shouldn't we fix that instead of adding a new flag to the error object?

The condition at the top of this file checks `err.shouldReportToSentry`, but this code is setting `err.reportToSentry`.

How can we (manually) verify that the changes in the pull request work as intended?

lib/zinnia.js
```diff
@@ -53,6 +53,9 @@ const maybeReportErrorToSentry = (/** @type {unknown} */ err) => {
   /** @type {Parameters<Sentry.captureException>[1]} */
   const hint = { extra: {} }
   if (typeof err === 'object') {
+    if ('shouldReportToSentry' in err && err.shouldReportToSentry === false) {
```
bajtos, line 56: An alternative name to consider: `reportedToSentry` or `wasReportedToSentry`.

Instead of setting this flag in various places where we want to report errors, maybe we can set the flag here, inside `maybeReportErrorToSentry`? That will prevent double-reports everywhere, not only in the place you modify in this PR.

```js
const maybeReportErrorToSentry = (/** @type {unknown} */ err) => {
  // ...
  if ('reportedToSentry' in err && err.reportedToSentry) return
  err.reportedToSentry = true
  // ...
}
```

bajtos (changes requested):

lib/zinnia.js
juliangruber: Screenshot 2024-04-29 at 16 15 10

Do you know how to fix this?

juliangruber:
I have changed the sentry key and disabled the old one

bajtos (approved):
I prefer slightly more `reportedToSentry` over `reportToSentry` for the following reasons:
- `reportedToSentry` describes what happened. When nobody reads/interprets this field, then it's ok - we are telling what happened, not asking what to do
- I understand `reportToSentry` as a command asking some other piece of code to report this error to Sentry. If that other piece of code does not report the error, then this command will be unfulfilled.

I also find it a bit weird when a property has a default value of `true`.

At the same time, this is subjective and a minor detail, so I am okay with landing this PR as it is now.

lib/zinnia.js
```diff
@@ -53,6 +53,10 @@ const maybeReportErrorToSentry = (/** @type {unknown} */ err) => {
   /** @type {Parameters<Sentry.captureException>[1]} */
   const hint = { extra: {} }
   if (typeof err === 'object') {
+    if ('reportToSentry' in err && err.reportToSentry === false) {
+      return
+    }
+    Object.assign(err, { reportToSentry: false })
```
juliangruber, line 59:
```suggestion
    if ('reportedToSentry' in err && err.reportToSentry === true) {
      return
    }
    Object.assign(err, { reportedToSentry: true })
```

lib/zinnia.js
```diff
@@ -53,6 +53,10 @@ const maybeReportErrorToSentry = (/** @type {unknown} */ err) => {
   /** @type {Parameters<Sentry.captureException>[1]} */
   const hint = { extra: {} }
   if (typeof err === 'object') {
+    if ('reportedToSentry' in err && err.reportToSentry === true) {
```
juliangruber, line 56:
```suggestion
    if ('reportedToSentry' in err && err.reportedToSentry === true) {
```

PR #437, opened 2024-04-30
Change Docker base image to slim

zipiju:
Slim shaves about 400MB off the container when installed. It is always better to have smaller images, especially when installing these at the edge on routers and such.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:20
+FROM node:20-slim
 LABEL org.opencontainers.image.source https://github.com/filecoin-station/core
 USER node
 WORKDIR /usr/src/app


juliangruber (approved):
+1 if CI passes

bajtos:
For context: we moved from Alpine to Full distro because Zinnia does not support Alpine yet and Saturn L2 module was not able to run on the Slim version.

See https://github.com/filecoin-station/core/pull/92
>  We cannot use `node:18-alpine` because Zinnia does not support that yet.
>
>  I tried to use `node:18-slim`, but Saturn L2 was not able to start in that system.
>  The Node.js project recommends to use `node:18`, so we should be fine. See https://github.com/nodejs/docker-node#image-variants

Now that we are no longer shipping Saturn L2, it makes a lot of sense to switch to Slim version.

Thank you for contributing this improvement, @zipiju! :heart:

PR #441, opened 2024-04-30
fix pass eth address to zinnia

juliangruber:
This way we only get 0x addresses submitted to `{spark,voyager}-api`


bajtos (approved):
:shipit:

PR #443, opened 2024-05-02
ci: add Node.js 22, drop 21

bajtos:
See https://nodejs.org/en/blog/announcements/v22-release-announce and https://twitter.com/_rafaelgss/status/1784932576924164227

> Node.js 21 is now in maintenance mode, which means this release line won't receive any further updates apart from security releases.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
-        node: ['18', '20', '21']
+        node: ['18', '20', '22']
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4


.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
-        node: ['18', '20', '21']
+        node: ['18', '20', '22']
```
juliangruber, line 15:
```suggestion
        node: ['18', '20', '21', '22']
```

I believe that dropping node 21 from CI is a breaking change (unless we consider node 21 itself unstable), because otherwise we can break node 21 support in the next patch release and won't notice it. We talked about odd numbered nodejs versions in another PR, but I can't quite recall the outcome 🤔

bajtos: I think the probability of breaking Node 21 only, while still running well on Node 20 and Node 22 is extremely low.

Having said that, I am fine running our CI on Node 21 as long as it's in maintenance mode.

bajtos:
@juliangruber LGTY now?

PR #462, opened 2024-05-14
move rewards update loop out of zinnia loop

juliangruber:
Slack discussions with context:
- https://space-meridian.slack.com/archives/C06RPCL6QGL/p1715686302232529
- https://filecoinproject.slack.com/archives/C023K7D9GAX/p1715419803449009

PR #463, opened 2024-05-14
refactor: code cleanup after rewards loop move

bajtos:
I am not entirely happy with the implementation details introduced in https://github.com/filecoin-station/core/pull/462. In particular, I find the `Obj` reference/pointer-like structure too unusual/non-idiomatic in JavaScript.

I am proposing the following two changes:

**refactor: allow submission of partial metrics**

Keep track of the last rewards & job count inside the metrics component so that we don't have to maintain & share that state between "lib/rewards" and "lib/zinnia".

**refactor: update `contracts` array in place**

We used to perform in-place updates before #462, this commit is effectively brining that old design back.

Also, remove `lib/obj.js` as it's no longer needed.


lib/metrics.js
```diff
@@ -28,10 +28,22 @@ export class Metrics {
    * - Writes `jobs-completed` to InfluxDB
    * - Merges metrics from all modules
    * @param {String} moduleName
-   * @param {MetricsEvent} metrics
+   * @param {Partial<MetricsEvent>} metrics
    */
   submit (moduleName, metrics) {
+    /** @type {MetricsEvent} */
+    const resolvedMetrics = {
```
juliangruber, line 35:
```suggestion
    const mergedMetrics = {
```

lib/metrics.js
```diff
@@ -44,7 +56,7 @@ export class Metrics {
           )
       )
     }
-    this.moduleMetrics.set(moduleName, metrics)
+    this.moduleMetrics.set(moduleName, resolvedMetrics)
```
juliangruber, line 59:
```suggestion
    this.moduleMetrics.set(moduleName, mergedMetrics)
```

juliangruber (approved):
Great ideas!

lib/metrics.js
bajtos: I cannot use the name `mergedMetrics` because we already have a variable with that name declared below.

```js
    const mergedMetrics = {
      totalJobsCompleted: 0,
      rewardsScheduledForAddress: 0n
    }
    for (const [, metrics] of this.moduleMetrics) {
      mergedMetrics.totalJobsCompleted += metrics.totalJobsCompleted
      // Merging rewards metrics should be revisited as more modules start
      // paying rewards
      mergedMetrics.rewardsScheduledForAddress +=
        metrics.rewardsScheduledForAddress
    }
```

bajtos: I am going to land this PR as it is now; feel free to open a follow-up PR to rename `resolvedMetrics` to something better.

PR #466, opened 2024-05-14
add validate eth addresses. closes #385

### LINKED ISSUE #385: Validate wallet addresses in 0x format
When the operator provides their FIL_WALLET_ADDRESS in `0x` format, we should validate it's a valid Ethereum/FEVM address and refuse to start if the address is not valid. The goal is to let the user know when they misconfigured their Station Core and won't receive any rewards.

Example of an invalid measurement we are receiving now:

```
Invalid measurement: valid participant address required Measurement {
2024-03-26T11:46:07Z app[178199dc20e638] cdg [info]  participantAddress: '0x6939091eb48F69057A2FEF00E69d88c0D06F47055',
```

juliangruber:
Closes #385

🧹

### DIFF
--- a/commands/station.js
+++ b/commands/station.js
@@ -8,7 +8,7 @@ import { paths } from '../lib/paths.js'
 import { getStationId } from '../lib/station-id.js'
 import pRetry from 'p-retry'
 import { fetch } from 'undici'
-import { ethAddressFromDelegated } from '@glif/filecoin-address'
+import { ethAddressFromDelegated, isEthAddress } from '@glif/filecoin-address'
 import { ethers, formatEther } from 'ethers'
 import { Obj } from '../lib/obj.js'
 import { runUpdateRewardsLoop } from '../lib/rewards.js'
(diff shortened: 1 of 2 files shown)


commands/station.js
```diff
@@ -44,6 +44,9 @@ export const station = async ({ json, experimental }) => {
   ) {
     panic('FIL_WALLET_ADDRESS must start with f410 or 0x')
   }
+  if (FIL_WALLET_ADDRESS.startsWith('0x') && !isEthAddress(FIL_WALLET_ADDRESS)) {
```
bajtos, line 47: We are converting f4 addresses to 0x format few lines below.

```js
  const ethAddress = FIL_WALLET_ADDRESS.startsWith('0x')
    ? FIL_WALLET_ADDRESS
    : ethAddressFromDelegated(FIL_WALLET_ADDRESS)
```

I propose to move the validation down there and validate the final `ethAddress` value that we pass to other components.

juliangruber: I thought about this too, but it's testing something else: It's testing whether the passed or the converted address is valid. As it is now, it's only testing whether the passed address is valid. I have a concern that the proposed method could shadow some tricky to find bugs.

bajtos (approved):
> I have a concern that the proposed method could shadow some tricky to find bugs.

Makes sense. Let's land the current version.

PR #474, opened 2024-05-22
fix: don't crash when getContractsWithRetry fails

bajtos:
See https://github.com/filecoin-station/desktop/issues/1547

### DIFF
--- a/lib/contracts.js
+++ b/lib/contracts.js
@@ -18,9 +18,13 @@ export const runUpdateContractsLoop = async ({ provider, abi, contracts }) => {
       if (err.name === 'AbortError') return
       throw err
     }
-    const newContracts = await getContractsWithRetry({ provider, abi })
-    contracts.splice(0)
-    contracts.push(...newContracts)
+    try {
+      const newContracts = await getContractsWithRetry({ provider, abi })
+      contracts.splice(0)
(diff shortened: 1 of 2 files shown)


lib/contracts.js
```diff
@@ -18,9 +18,13 @@ export const runUpdateContractsLoop = async ({ provider, abi, contracts }) => {
...
     }
-    const newContracts = await getContractsWithRetry({ provider, abi })
-    contracts.splice(0)
-    contracts.push(...newContracts)
+    try {
+      const newContracts = await getContractsWithRetry({ provider, abi })
+      contracts.splice(0)
+      contracts.push(...newContracts)
+    } catch (err) {
+      console.error('Failed to update the list of contract addresses. Will retry later.', err)
```
juliangruber, line 26: What do you think about creating an activity entry here? Say this always fails, then the user should know.

juliangruber (changes requested):

lib/contracts.js
bajtos: See [14f93ab](https://github.com/filecoin-station/core/pull/474/commits/14f93ab87dc1c84088218ac7fabe5805b10429ae)

juliangruber (approved):

PR #477, opened 2024-05-22
contracts: update wait after fetch

juliangruber:
Now that the contract update loop is decoupled from modules crashing, we can fetch updates immediately without needing to worry about ddosing w3name

### DIFF
--- a/lib/contracts.js
+++ b/lib/contracts.js
@@ -10,6 +10,10 @@ const {
 
 export const runUpdateContractsLoop = async ({ provider, abi, contracts }) => {
   while (true) {
+    const newContracts = await getContractsWithRetry({ provider, abi })
+    contracts.splice(0)
+    contracts.push(...newContracts)
+
     const delay = 10 * 60 * 1000 // 10 minutes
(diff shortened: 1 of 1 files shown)


lib/contracts.js
```diff
@@ -10,6 +10,10 @@ const {
 
 export const runUpdateContractsLoop = async ({ provider, abi, contracts }) => {
   while (true) {
```
bajtos, line 12: Should we add a small delay (e.g. 2 seconds) before we start the loop to avoid DDoSing w3name in case Station Core is frequently restarted? For example, if somebody is operating Station Core and has a mechanism to restart Core immediately after it exits.

```suggestion
export const runUpdateContractsLoop = async ({ provider, abi, contracts }) => {
  await timers.setTimeout(2_000)
  while (true) {
```

bajtos (approved):

lib/contracts.js
juliangruber: great idea! 👏

bajtos (approved):
:shipit:

PR #551, opened 2024-07-24
fix redundant exit message

juliangruber:
Remove the `exit` event handler which would create duplicate activity events on a crash.

### DIFF
--- a/lib/zinnia.js
+++ b/lib/zinnia.js
@@ -163,6 +163,7 @@ const catchChildProcessExit = async ({
     const tasks = childProcesses.map(p => (async () => {
       try {
         await p
+        onActivity({ type: 'info', message: `${capitalize(module)} exited` })
       } catch (err) {
         // When the child process crash, attach the module name & the exit reason to the error object
         const exitReason = p.exitCode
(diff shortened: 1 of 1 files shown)


lib/zinnia.js
```diff
@@ -294,12 +295,6 @@ export async function run ({
       resetTimeout()
       process.stderr.write(data)
     })
-
-    childProcess.on('exit', (code, signal) => {
-      const exitReason = signal ? `via signal ${signal}` : `with code: ${code}`
-      const msg = `${capitalize(module)} exited ${exitReason}`
```
bajtos, line 300: We are losing the information on whether the module exited via a signal or by returning a (potentially non-zero) exit code. Can we preserve it, please?

bajtos: Oh, I see. This code part of our codebase is truly difficult to understand 😢

bajtos (approved):
IIUC, your change is removing redundant activity log entries printed when the Zinnia process exits either with a non-zero exit code or via a signal. After your change, only one activity should be logged. That's a great improvement! 👏🏻

lib/zinnia.js
juliangruber: Yeah, that's I think why this was unnoticed for a while

juliangruber: This is taking a step in the right direction, by removing some branches :)

PR #563, opened 2024-08-01
Add `--recreateStationIdOnError`

juliangruber:
For https://github.com/filecoin-station/desktop/issues/1726

I decided to add this flag instead of a new exit code (as suggested in https://github.com/filecoin-station/desktop/issues/1726#issuecomment-2262594686), because with the new exit code Station Desktop needs to know where Station Core stores its internal files, and this coupling isn't necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,10 +140,12 @@ $ station --help
 Usage: station [options]
 
 Options:
-  -j, --json          Output JSON                                      [boolean]
-      --experimental  Also run experimental modules                    [boolean]
-  -v, --version       Show version number                              [boolean]
-  -h, --help          Show help                                        [boolean]
+  -j, --json                      Output JSON                          [boolean]
+      --experimental              Also run experimental modules        [boolean]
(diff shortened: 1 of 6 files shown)


juliangruber:
Blocks https://github.com/filecoin-station/desktop/pull/1742

bajtos (changes requested):
The proposed implementation LGTM.

Let's add some test please.

- https://github.com/filecoin-station/core/blob/main/test/cli.js
- https://github.com/filecoin-station/core/blob/main/test/station-id.test.js

bajtos (approved):
👌🏻 

What do you think about improving the following error message to explain to the user that they can re-run Station Core with the new flag to fix the problem?

https://github.com/filecoin-station/core/blob/fe02c6e4e10f043743a640c6fc807c4af8f76c8f/lib/station-id.js#L64

juliangruber:
Great idea!

PR #582, opened 2024-08-27
zinnia: remove logging retried operations to Sentry

juliangruber:
When the overall operation failed it will still be logged.


bajtos (approved):
👍🏻

PR #594, opened 2024-09-10
Rotate Sentry DSN

juliangruber:


bajtos:
FWIW, I have already rotated the token last week and disabled the old one.
See https://github.com/filecoin-station/core/pull/588

bajtos:
I disabled the old token only yesterday, that's why the old error is still present in the weekly summary.

bajtos (approved):
Having said that, there is nothing wrong with this PR, we can land it if you like.

juliangruber:
got it! Then this doesn't need to be released now

PR #606, opened 2024-10-07
Read scheduled rewards from `spark-rewards`

juliangruber:
https://github.com/space-meridian/roadmap/issues/161

### DIFF
--- a/lib/rewards.js
+++ b/lib/rewards.js
@@ -15,7 +15,7 @@ export const runUpdateRewardsLoop = async ({ contracts, ethAddress, onMetrics })
       ...contracts.map(async contract => {
         return getContractScheduledRewardsWithFallback(contract, ethAddress)
       }),
-      getSparkRewardsScheduledRewardsWithFallback(ethAddress)
+      getOffchainScheduledRewardsWithFallback(ethAddress)
     ])
     const totalRewards = rewards.reduce((a, b) => a + b, 0n)
     onMetrics({ rewardsScheduledForAddress: totalRewards })
@@ -26,7 +26,7 @@ export const runUpdateRewardsLoop = async ({ contracts, ethAddress, onMetrics })
   }
 }
 
-async function getSparkRewardsScheduledRewardsWithFallback (ethAddress) {
+async function getOffchainScheduledRewardsWithFallback (ethAddress) {
   try {
     const res = await fetch(
       `https://spark-rewards.fly.dev/scheduled-rewards/${ethAddress}`
--- a/lib/rewards.js
+++ b/lib/rewards.js
@@ -1,4 +1,5 @@
 import timers from 'node:timers/promises'
+import assert from 'node:assert/strict'
 
 /**
  * @param {object} args
@@ -32,9 +33,8 @@ async function getOffchainScheduledRewardsWithFallback (ethAddress) {
       `https://spark-rewards.fly.dev/scheduled-rewards/${ethAddress}`
     )
     const json = await res.json()
-    return typeof json === 'string'
-      ? BigInt(json)
-      : 0n // json can be `null`
+    assert(typeof json === 'string')
+    return BigInt(json)
   } catch (err) {
     console.error('Failed to get scheduled rewards:', err.stack)
     return 0n
--- a/lib/rewards.js
+++ b/lib/rewards.js
@@ -11,10 +11,13 @@ export const runUpdateRewardsLoop = async ({ contracts, ethAddress, onMetrics })
     while (!contracts.length) {
       await timers.setTimeout(1000)
     }
-    const contractRewards = await Promise.all(contracts.map(async contract => {
-      return getScheduledRewardsWithFallback(contract, ethAddress)
-    }))
-    const totalRewards = contractRewards.reduce((a, b) => a + b, 0n)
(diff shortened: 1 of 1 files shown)


lib/rewards.js
```diff
@@ -11,10 +11,13 @@ export const runUpdateRewardsLoop = async ({ contracts, ethAddress, onMetrics })
...
     }
-    const contractRewards = await Promise.all(contracts.map(async contract => {
-      return getScheduledRewardsWithFallback(contract, ethAddress)
-    }))
-    const totalRewards = contractRewards.reduce((a, b) => a + b, 0n)
+    const rewards = await Promise.all([
+      ...contracts.map(async contract => {
+        return getContractScheduledRewardsWithFallback(contract, ethAddress)
+      }),
+      getSparkRewardsScheduledRewardsWithFallback(ethAddress)
```
juliangruber, line 18: This name is awkward but correct. @bajtos do you have a better suggestion?

lib/rewards.js
```diff
@@ -23,7 +26,22 @@ export const runUpdateRewardsLoop = async ({ contracts, ethAddress, onMetrics })
...
-async function getScheduledRewardsWithFallback (contract, ethAddress) {
+async function getSparkRewardsScheduledRewardsWithFallback (ethAddress) {
+  try {
+    const res = await fetch(
+      `https://spark-rewards.fly.dev/scheduled-rewards/${ethAddress}`
+    )
+    const json = await res.json()
+    return typeof json === 'string'
+      ? BigInt(json)
+      : 0n // json can be `null`
```
juliangruber, line 37: I was thinking about changing the service to return `"0"` also when the participant isn't known. This matches the behavior of the smart contract. Wdyt @bajtos?

lib/rewards.js
bajtos: How about `getOffchainScheduledRewardsWithFallback`?

lib/rewards.js
bajtos: I don't have a strong opinion.

On one hand, as a REST API service client, I would expect the service to return 404 for entities (or URL paths) that don't exist.

On the other hand, a path with an address we haven't seen before is a valid entity, plus it makes the implementation on the client side easier, so it makes sense to return `0` in such case.

lib/rewards.js
```diff
@@ -23,7 +26,22 @@ export const runUpdateRewardsLoop = async ({ contracts, ethAddress, onMetrics })
   }
 }
 
-async function getScheduledRewardsWithFallback (contract, ethAddress) {
+async function getSparkRewardsScheduledRewardsWithFallback (ethAddress) {
+  try {
+    const res = await fetch(
+      `https://spark-rewards.fly.dev/scheduled-rewards/${ethAddress}`
+    )
+    const json = await res.json()
```
bajtos, line 34: How do you handle 4xx and 5xx responses? What if the server responds with 5xx and the response body provides error details? Should you check `res.ok`?

In other places, I had a great experience with the helper `assertOkResponse` that reads the error response body and includes it in `err.message`.

bajtos (approved):

lib/rewards.js
juliangruber: > How do you handle 4xx and 5xx responses? What if the server responds with 5xx and the response body provides error details? Should you check res.ok?

I believe we can get to these what ifs when we need. For now, we know that it will throw when there's a bad response, and it will contain the error:

```
fetch('https://ohnothisdoesnotexist.beep').catch(console.error)
TypeError: fetch failed
    at node:internal/deps/undici/undici:13178:13
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  [cause]: Error: getaddrinfo ENOTFOUND ohnothisdoesnotexist.beep
      at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:120:26)
      at GetAddrInfoReqWrap.callbackTrampoline (node:internal/async_hooks:130:17) {
    errno: -3008,
    code: 'ENOTFOUND',
    syscall: 'getaddrinfo',
    hostname: 'ohnothisdoesnotexist.beep'
  }
}
```

`spark-rewards` does not return a JSON error in this route, so parsing the response is premature. We could include the response text in the error, but this route also doesn't do that - if an error happens, it returns `Internal Server Error`.

juliangruber:
Blocked by https://github.com/filecoin-station/spark-rewards/issues/13

lib/rewards.js
bajtos: FWIW:
> ```js
> fetch('https://ohnothisdoesnotexist.beep').catch(console.error)
> ```

That's a different error case. This is what I had in mind:

```
(await fetch('https://example.com')).json().catch(console.error)
SyntaxError: Unexpected token '<', "<!doctype "... is not valid JSON
    at JSON.parse (<anonymous>)
    at parseJSONFromBytes (node:internal/deps/undici/undici:5329:19)
    at successSteps (node:internal/deps/undici/undici:5300:27)
    at fullyReadBody (node:internal/deps/undici/undici:1447:9)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async specConsumeBody (node:internal/deps/undici/undici:5309:7)
```

PR #623, opened 2024-11-10
modules: ensure main file also exists

juliangruber:
Trying to fix reported cases of `sources/spark/main.js` missing, while `sources/spark` is there.

This is the same path we're checking below - only if the main file `main.js` exists, consider the module as functional. This is still not great but should fix some cases. In the future, use hashes to check validity of _all_ files on disk.


bajtos (commented):
👏🏻

PR #671, opened 2025-01-30
Fix docker after moving to `CheckerNetwork`

juliangruber:
This prevents new users from installing from the old URL. Independent of this, I will notify existing users of the new URL.


bajtos (approved):
:shipit:

PR #679, opened 2025-02-10
zinnia: fix reporting AbortErrors to Sentry

juliangruber:
Fixes https://space-meridian.sentry.io/issues/5857510005/?query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=freq&statsPeriod=14d&stream_index=1 and more.

Before this change, an abort error will always be reported to Sentry. Situations that trigger an abort:

- Module has been inactive (I have added error handling specifically for this here)
- Any other error happened (this is already being reported to Sentry, no need to report the resulting abort error)
- Restarting Zinnia after updating source files (not an error)

Looking at Sentry, and especially the view of all issues sorted by number of events, this is the biggest offender that is still happening.

### DIFF
--- a/lib/zinnia.js
+++ b/lib/zinnia.js
@@ -319,8 +319,12 @@ export async function run ({
     ])
     console.error('Zinnia main loop ended')
   } catch (err) {
-    console.error('Zinnia main loop errored', err)
-    maybeReportErrorToSentry(err)
+    if (err.name === 'AbortError') {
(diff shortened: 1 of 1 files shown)


bajtos (approved):
Thank you for looking into this issue!

juliangruber:
https://github.com/CheckerNetwork/core/releases/tag/v21.2.6

PR #710, opened 2025-04-02
chore: configure neostandard + prettier

bajtos:
I copied eslint and prettier configuration from https://github.com/CheckerNetwork/assert-ok-response

I split the changes into two commits to make this pull request easier to review. The second commit contains linting fixes automatically generated by eslint and prettier.

Last but not least, I also reworked npm scripts for linting to be consistent with other repositories.

- **chore: configure neostandard + prettier**
- **chore: `npm run lint:fix`**

Links:
- https://github.com/CheckerNetwork/roadmap/issues/166

### DIFF
--- a/.prettierrc.yaml
+++ b/.prettierrc.yaml
@@ -1,2 +1,0 @@
-semi: false
-singleQuote: true
--- a/.prettierrc.yaml
+++ b/.prettierrc.yaml
@@ -1,2 +1,5 @@
-printWidth: 80
+semi: false
+singleQuote: true
+printWidth: 100
+trailingComma: all
 proseWrap: always
--- a/.prettierrc.yaml
+++ b/.prettierrc.yaml
@@ -1,5 +1,2 @@
 semi: false
 singleQuote: true
-printWidth: 100
-trailingComma: all
-proseWrap: always
--- a/commands/checker.js
+++ b/commands/checker.js
@@ -13,14 +13,9 @@ import { ethers, formatEther } from 'ethers'
 import { runUpdateRewardsLoop } from '../lib/rewards.js'
 import { runUpdateContractsLoop } from '../lib/contracts.js'
 
-const {
-  FIL_WALLET_ADDRESS,
-  PASSPHRASE
-} = process.env
+const { FIL_WALLET_ADDRESS, PASSPHRASE } = process.env
 
-const runtimeNames = [
-  'zinnia'
-]
+const runtimeNames = ['zinnia']
 
 /**
  * @param {string} msg
@@ -31,10 +26,16 @@ const panic = (msg, exitCode = 1) => {
   process.exit(exitCode)
 }
 
-export const checker = async ({ json, recreateCheckerIdOnError, experimental }) => {
+export const checker = async ({
+  json,
+  recreateCheckerIdOnError,
+  experimental,
+}) => {
   if (!FIL_WALLET_ADDRESS) panic('FIL_WALLET_ADDRESS required')
   if (FIL_WALLET_ADDRESS.startsWith('f1')) {
-    panic('Invalid FIL_WALLET_ADDRESS: f1 addresses are currently not supported. Please use an f4 or 0x address.')
+    panic(
+      'Invalid FIL_WALLET_ADDRESS: f1 addresses are currently not supported. Please use an f4 or 0x address.',
+    )
   }
   if (
     !FIL_WALLET_ADDRESS.startsWith('f410') &&
@@ -42,20 +43,30 @@ export const checker = async ({ json, recreateCheckerIdOnError, experimental }) 
   ) {
     panic('FIL_WALLET_ADDRESS must start with f410 or 0x')
   }
-  if (FIL_WALLET_ADDRESS.startsWith('0x') && !isEthAddress(FIL_WALLET_ADDRESS)) {
+  if (
+    FIL_WALLET_ADDRESS.startsWith('0x') &&
+    !isEthAddress(FIL_WALLET_ADDRESS)
+  ) {
     panic('Invalid FIL_WALLET_ADDRESS ethereum address', 2)
   }
 
-  const keypair = await getCheckerId({ secretsDir: paths.secrets, passphrase: PASSPHRASE, recreateOnError: recreateCheckerIdOnError })
+  const keypair = await getCheckerId({
+    secretsDir: paths.secrets,
+    passphrase: PASSPHRASE,
+    recreateOnError: recreateCheckerIdOnError,
+  })
   const CHECKER_ID = keypair.publicKey
 
   const fetchRes = await pRetry(
-    () => fetch(`https://station-wallet-screening.fly.dev/${FIL_WALLET_ADDRESS}`),
+    () =>
+      fetch(`https://station-wallet-screening.fly.dev/${FIL_WALLET_ADDRESS}`),
     {
       retries: 1000,
       onFailedAttempt: () =>
-        console.error('Failed to validate FIL_WALLET_ADDRESS address. Retrying...')
-    }
+        console.error(
+          'Failed to validate FIL_WALLET_ADDRESS address. Retrying...',
+        ),
+    },
   )
   if (fetchRes.status === 403) panic('Invalid FIL_WALLET_ADDRESS address', 2)
   if (!fetchRes.ok) panic('Failed to validate FIL_WALLET_ADDRESS address')
@@ -67,42 +78,55 @@ export const checker = async ({ json, recreateCheckerIdOnError, experimental }) 
     await fs.mkdir(join(paths.runtimeState, runtimeName), { recursive: true })
   }
 
-  activities.onActivity(activity => {
+  activities.onActivity((activity) => {
     if (json) {
-      console.log(JSON.stringify({
-        type: `activity:${activity.type}`,
-        subnet: activity.source,
-        message: activity.message
-      }))
+      console.log(
+        JSON.stringify({
+          type: `activity:${activity.type}`,
+          subnet: activity.source,
+          message: activity.message,
+        }),
+      )
     } else {
       process.stdout.write(formatActivityObject(activity))
     }
   })
 
-  metrics.onUpdate(metrics => {
+  metrics.onUpdate((metrics) => {
     if (json) {
-      console.log(JSON.stringify({
-        type: 'jobs-completed',
-        total: metrics.totalJobsCompleted,
-        rewardsScheduledForAddress: formatEther(metrics.rewardsScheduledForAddress)
-      }))
+      console.log(
+        JSON.stringify({
+          type: 'jobs-completed',
+          total: metrics.totalJobsCompleted,
+          rewardsScheduledForAddress: formatEther(
+            metrics.rewardsScheduledForAddress,
+          ),
+        }),
+      )
     } else {
-      console.log(JSON.stringify({
-        totalJobsCompleted: metrics.totalJobsCompleted,
-        rewardsScheduledForAddress:
-          formatEther(metrics.rewardsScheduledForAddress)
-      }, null, 2))
+      console.log(
+        JSON.stringify(
+          {
+            totalJobsCompleted: metrics.totalJobsCompleted,
+            rewardsScheduledForAddress: formatEther(
+              metrics.rewardsScheduledForAddress,
+            ),
+          },
+          null,
+          2,
+        ),
+      )
     }
   })
 
   const contracts = []
 
   const fetchRequest = new ethers.FetchRequest(
-    'https://api.node.glif.io/rpc/v1'
+    'https://api.node.glif.io/rpc/v1',
   )
   fetchRequest.setHeader(
     'Authorization',
-    'Bearer RXQ2SKH/BVuwN7wisZh3b5uXStGPj1JQIrIWD+rxF0Y='
+    'Bearer RXQ2SKH/BVuwN7wisZh3b5uXStGPj1JQIrIWD+rxF0Y=',
   )
   const provider = new ethers.JsonRpcProvider(fetchRequest)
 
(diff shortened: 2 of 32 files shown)


.prettierrc.yaml
```diff
@@ -1,2 +1,5 @@
-printWidth: 80
+semi: false
+singleQuote: true
+printWidth: 100
```
juliangruber, line 3: The PR this was copied from didn't get reviewed, therefore I believe here is the place where we need to discuss standards that work for the team. @bajtos can you please manage that process?

Things I would like to change:
- printWidth: **80** (tradition, enforces low nesting)
- trailingComma: **none** (I personally really don't like trailing commas)

As a compromise as always, I like using tools without changing their configuration. So, if we can just use prettier defaults, I'll +1 it (even though it uses trailing commas). This will help keep our code base opinion free, and follow community standards.

juliangruber (changes requested):

.prettierrc.yaml
juliangruber: I realize I had already approved https://github.com/CheckerNetwork/spark-checker/pull/127. This was for one repo though. If we copy this exact config to every repo, I would like to have a discussion about it first please.

bajtos: SGTM, let's discuss the desired style here 👍🏻 

**trailing comma**

Some prior discussions happened in https://github.com/CheckerNetwork/roadmap/issues/166

> > [bajtos] Most importantly, Standard does not allow trailing commas in arrays & object literals. I am frustrated by this several times a day. https://github.com/standard/standard/issues/240
>
> [juliangruber] When it comes to form, I dislike them. When it comes to function, they are keeping git diffs clean, so I'm in favor.

FWIW, `trailingComma: all` is the current default behaviour of Prettier. (The default value changed in Prettier v3.0.0.) See https://prettier.io/docs/options#trailing-commas

**text width**

See https://prettier.io/docs/options#print-width

- The default is 80 characters, as you suggested, @juliangruber 
- Prettier recommends against using more than 80 characters (see the doc page I linked to)

.prettierrc.yaml
```diff
@@ -1,2 +1,5 @@
-printWidth: 80
+semi: false
```
bajtos, line 1: Should we keep using no semicolons or should we switch to Prettier defaults and start using semicolons again?

https://prettier.io/docs/options#semicolons

I slightly prefer to use semicolons for consistency with Rust, where a statement with a semicolon has a different meaning than a statement without one.

I believe the default TypeScript formatter uses semicolons too. But they also use double quotes, IIRC, while we are using single quotes, so this may not be the best example.

.prettierrc.yaml
```diff
@@ -1,2 +1,5 @@
...
+singleQuote: true
```
bajtos, line 2: Do we prefer single quotes or should we switch to Prettier defaults - double quotes?

https://prettier.io/docs/options#quotes

.prettierrc.yaml
```diff
@@ -1,2 +1,5 @@
...
+printWidth: 100
+trailingComma: all
```
bajtos, line 4: `trailingComma: all` is Prettier's default, we are discussing this in https://github.com/CheckerNetwork/node/pull/710/files#r2024437202

.prettierrc.yaml
```diff
@@ -1,2 +1,5 @@
...
 proseWrap: always
```
bajtos, line 5: How do we want to wrap Markdown text?

https://prettier.io/docs/options#prose-wrap

I believe we are already using `proseWrap: always` in all our repositories.

By default, Prettier does not apply any line wrapping rules for Markdown text.

.prettierrc.yaml
bajtos: My preference: use single quotes

.prettierrc.yaml
bajtos: My preference: keep `proseWrap: always`.

Long lines are difficult to review on GitHub, as a comment for one line applies to the entire paragraph on that single line.

.prettierrc.yaml
juliangruber: Cool, I'm happy with default trailing comma and default max width

.prettierrc.yaml
juliangruber: My preference is without semicolons

.prettierrc.yaml
pyropy: I really like trailing commas👉🏻👈🏻 

Other precerences: single quotes, shorther print width (80 seems okay) to avoid nesting.

.prettierrc.yaml
juliangruber: I find diffs very noisy with wrapping, and prefer without wrapping

.prettierrc.yaml
pyropy: I would vote in favor of no semicolons. Personally I consider code with no semicolons more readable.

bajtos:
I updated the Prettier config based on our discussions above, removed settings where we use Prettier defaults and kept only settings where we differ.

The new Prettier config:

```yaml
semi: false
singleQuote: true
```

After reviewing the changes, I added two more files to `.prettierignore`:

- `/lib/abi.json` - this is an auto-generated definition of a smart-contract ABI we copy from elsewhere, IIRC.
- `/tsconfig.json` - let's keep the list of included/excluded files formatted as one entry per line to reduce noisy diffs in the future.

I consider this PR ready for final review & merging.

juliangruber:
> /tsconfig.json - let's keep the list of included/excluded files formatted as one entry per line to reduce noisy diffs in the future.

Can we instead //ignore that one line, and lint the rest? I don't like this special case. You can make the same argument for other files with arrays. Instead, should we tweak the config to always list array elements on multiple lines?

bajtos:
> > /tsconfig.json - let's keep the list of included/excluded files formatted as one entry per line to reduce noisy diffs in the future.
> 
> Can we instead //ignore that one line, and lint the rest? I don't like this special case. You can make the same argument for other files with arrays. Instead, should we tweak the config to always list array elements on multiple lines?

To my best knowledge:
- You cannot add ignore directives to JSON files, because JSON does not support comments.
- I could not find any config option to tell Prettier to use one line per array item in JSON files

I see two conflicting requirements here:
- We want to avoid noisy diffs
- We don't want special cases

IIUC, avoiding special cases is more important to you than avoiding noisy diffs. Is that correct?

I don't have a strong opinion myself, I am going to remove `tsconfig.json` from `.prettierignore`

bajtos:
> I am going to remove `tsconfig.json` from `.prettierignore`


See [9c67ebb](https://github.com/CheckerNetwork/node/pull/710/commits/9c67ebbc8f0fe4981262bc39072edb9d6904192b)

.prettierrc.yaml
NikolasHaimerl: I am with Julian on this one. Using the default settings usually gives you the experience the creator of a tooling had in mind and that usually works well. It's also the easiest to keep consistent across repos.

juliangruber:
Have you looked into the other option I suggested, @bajtos? To configure the linter to always multi-line arrays, so that we avoid all of these cases?

bajtos:
> Have you looked into the other option I suggested, @bajtos? To configure the linter to always multi-line arrays, so that we avoid all of these cases?

Yes, I did. Quoting from https://github.com/CheckerNetwork/node/pull/710#issuecomment-2777709367, with added emphasis:

> To my best knowledge:
> 
> * You cannot add ignore directives to JSON files, because JSON does not support comments.
> * **I could not find any config option to tell Prettier to use one line per array item in JSON files**

You can find the list of all Prettier options here:

https://prettier.io/docs/options

juliangruber:
Oh sorry I missed that. 😅 

There's https://www.npmjs.com/package/prettier-plugin-multiline-arrays as an option, if you agree I'm happy to add it to this PR.

Btw, I think we should soon think about bundling our linting logic into a module, so we can share the configuration among all our repos, and get dependabot updates for them. I'm also happy to work on that

bajtos:
> Oh sorry I missed that. 😅
> 
> There's https://www.npmjs.com/package/prettier-plugin-multiline-arrays as an option, if you agree I'm happy to add it to this PR.

Nice find!

If we are going to add prettier plugins, how do you feel about adding the following plugins too? (Discovered via https://github.com/electrovir/prettier-plugin-multiline-arrays?tab=readme-ov-file#compatibility):

- https://www.npmjs.com/package/prettier-plugin-packagejson
- https://www.npmjs.com/package/prettier-plugin-organize-imports
- prettier-plugin-jsdoc

> Btw, I think we should soon think about bundling our linting logic into a module, so we can share the configuration among all our repos, and get dependabot updates for them. I'm also happy to work on that

Great idea! I propose the following next steps:

1. Finish the discussion about the desired Neostandard & Prettier config in this pull request.
2. Land this PR.
3. Create a shared configuration
4. Update this repo to use the new shared configuration
5. Continue updating other repositories.

I propose creating two repositories for the shared config:
- CheckerNetwork/eslint-config
- CheckerNetwork/prettier-config

Thoughts?

juliangruber:
> If we are going to add prettier plugins, how do you feel about adding the following plugins too? (Discovered via https://github.com/electrovir/prettier-plugin-multiline-arrays?tab=readme-ov-file#compatibility):

> https://www.npmjs.com/package/prettier-plugin-packagejson

+1

> https://www.npmjs.com/package/prettier-plugin-organize-imports

+100

> prettier-plugin-jsdoc

https://www.npmjs.com/package/prettier-plugin-jsdoc

+100



> > Btw, I think we should soon think about bundling our linting logic into a module, so we can share the configuration among all our repos, and get dependabot updates for them. I'm also happy to work on that
> 
> Great idea! I propose the following next steps:
> 
> 1. Finish the discussion about the desired Neostandard & Prettier config in this pull request.
> 2. Land this PR.
> 3. Create a shared configuration
> 4. Update this repo to use the new shared configuration
> 5. Continue updating other repositories.

💯 

> I propose creating two repositories for the shared config:
> 
> * CheckerNetwork/eslint-config
> * CheckerNetwork/prettier-config
> 
> Thoughts?

I had assumed we can create a binary, so it's just

```json
{
  "dependencies": {
    "@space-meridian/lint": "0.0.0"
  },
  "scripts": {
    "test:lint": "space-lint"
  }
}
```

This way, the project doesn't need to bother and maintained tool versions (prettier & eslint). I do wonder about whether we need extensibility of the configuration, and therefore the approach you suggested works better. Wdyt?

bajtos:
> I had assumed we can create a binary

Unfortunately, that does not work with IDEs like VSCode and Zed. There was a long discussion between Neostandard and Eslint maintainers on this topic, see e.g. https://github.com/neostandard/neostandard/pull/2, https://github.com/neostandard/neostandard/issues/33 and other linked issues/PRs.

bajtos:
> https://www.npmjs.com/package/prettier-plugin-organize-imports
>
> +100

I am not able to get this working. Let's leave it up for future iterations.

bajtos:
I added a bunch of commits to configure additional Prettier plugins:
- prettier-plugin-multiline-arrays
- prettier-plugin-packagejson
- prettier-plugin-jsdoc

For each plugin, there are two commits - one with the configuration and another with the formatting changes. I hope it will make it easier to review this pull request.

juliangruber:
> I added a bunch of commits to configure additional Prettier plugins:
> 
> * prettier-plugin-multiline-arrays
> * prettier-plugin-packagejson
> * prettier-plugin-jsdoc
> 
> For each plugin, there are two commits - one with the configuration and another with the formatting changes. I hope it will make it easier to review this pull request.

FYI, for me, I usually don't review individual commits, as you can't add comments. With the full PR view some things are a bit harder to see, but I prefer that approach for any PR that isn't thousands of lines.

commands/checker.js
```diff
@@ -13,14 +13,9 @@ import { ethers, formatEther } from 'ethers'
...
-const {
-  FIL_WALLET_ADDRESS,
-  PASSPHRASE
-} = process.env
+const { FIL_WALLET_ADDRESS, PASSPHRASE } = process.env
 
-const runtimeNames = [
-  'zinnia'
-]
+const runtimeNames = ['zinnia']
```
juliangruber, line 18: Interesting, didn't we want things like this to be multi line arrays?

juliangruber: Do we need to set the threshold to 0?

lib/metrics.js
```diff
@@ -41,65 +42,71 @@ export class Metrics {
...
     const mergedMetrics = {
       totalJobsCompleted: 0,
-      rewardsScheduledForAddress: 0n
+      rewardsScheduledForAddress: 0n,
     }
-    for (const [, metrics] of this.subnetMetrics) {
+    for (const [
+      ,
+      metrics,
+    ] of this.subnetMetrics) {
```
juliangruber, line 75: This looks quite bad. Restructuring with multiline arrays is pretty ugly :|

lib/metrics.js
```diff
@@ -41,65 +42,71 @@ export class Metrics {
...
-      Object
-        .entries(this.mergedMetrics)
-        .some(([key, value]) => mergedMetrics[key] !== value)
+    const isChanged =
+      this.mergedMetrics === null ||
+      Object.entries(this.mergedMetrics).some(
+        ([
+          key,
+          value,
+        ]) => mergedMetrics[key] !== value,
```
juliangruber, line 88: Same here wrt destructuring and multi line arrays

juliangruber: Maybe this plugin won't work so well after all, and maybe that's why it's not included in prettier core

scripts/version.js
```diff
@@ -7,4 +7,7 @@ import * as paths from '../lib/paths.js'
 const pkg = JSON.parse(await fs.readFile(paths.packageJSON, 'utf8'))
 pkg.sentryEnvironment = 'production'
 await fs.writeFile(paths.packageJSON, JSON.stringify(pkg, null, 2) + '\n')
-await execa('git', ['add', 'package.json'])
+await execa('git', [
+  'add',
+  'package.json',
+])
```
juliangruber, line 13: Reading this PR more and more, I think I would like to remove the multi line array plugin. The desire to keep diffs clean is imposing a too inconvenient style

commands/checker.js
bajtos: I tried to set the threshold to `0`, but that made other lines look ugly. I will add a commit to show the result.

bajtos: Reverted in https://github.com/CheckerNetwork/node/pull/710/commits/1c1ff2b0e751545587c8cfbb14a11447d112f16d

commands/checker.js
```diff
@@ -67,42 +78,55 @@ export const checker = async ({ json, recreateCheckerIdOnError, experimental })
...
+          {
+            totalJobsCompleted: metrics.totalJobsCompleted,
+            rewardsScheduledForAddress: formatEther(
+              metrics.rewardsScheduledForAddress,
+            ),
+          },
+          null,
+          2,
+        ),
+      )
```
bajtos, line 118: This would be better formatted if we used `%o` instead of JSON.stringify, I think.

```js
console.log('%o', {
 totalJobsCompleted: metrics.totalJobsCompleted,
 rewardsScheduledForAddress: formatEther(
 metrics.rewardsScheduledForAddress,
})
```

I feel such a change is out of the scope of this PR. WDYT?

pyropy (approved):
Trailing commas woo-hoo! 🎉 

Overall this looks very good to me. I don't have any strong opinions on how the code is formatted now but I'm glad we have settled on styling configuration. 👍🏻

juliangruber (approved):
love it

PR #732, opened 2025-04-24
ci: auto-approve only minor prettier-config updates

bajtos:
This is a follow-up for eef0e609 (#726), which enabled auto-approvals for _all_ new versions of `@checkernetwork/prettier-config`, and the discussion in https://github.com/CheckerNetwork/spark-checker/pull/133#discussion_r2055795677 where we decided to auto-approve only minor updates.


juliangruber (approved):

juliangruber:
@bajtos I realized another reason for my arguing: While some packages behave semver-badly and intentionally release breaking changes in minor updates, those should be the exception and not the norm. The norm should be that anything that is a patch should always be merged, anything that is a minor should be merged + the devs notified, in case they want to use the new feature, anything that is a major needs careful review.

